PR #50, opened 2021-04-22
Fix minor bugs and added changes relating to PR 42

ssha0054:
# Description 
- This PR fixes minor bugs in the backend relating to the response codes used. It also includes a new GET request which allows us to retrieve a specific user by providing their id.

# Type of change 
- [x] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
Not the most appropriate response codes were used. Also, missing functionality to retrieve a specific user by their ID was not included.

# Solution description

- Replaced some of the previous response codes used with new ones that are more appropriate.
- Added a new GET request for the end point 'users/:id', which allows us to retrieve a specific user by their id provided by mongo.

# Tests 
- Tested new GET request for user via ID using insomnia.
image

Steps to reproduce:
1. Do a GET request on 'http://localhost:5000/users/' to show all users.
2. Copy the id of a particular user.
3. Do a GET request on 'http://localhost:5000/users/id' where 'id' is the copied id of a user.
4. Only the user with that id should be returned. 

Reference to response codes can be found on:
https://restfulapi.net/http-status-codes/ 

# Risk
Low


kdav108 (approved):
Thanks for this change!

PR #58, opened 2021-04-24
Update changes in `main` from `Logiciel` into `Thunder`

kdav108:
# Description 
This PR brings changes in `main` into `thunder` and in doing so integrates changes of Thunder with those in `main` so the system is integrated as a whole.

I know there's a lot of files changed shown on GitHub, but most are already reviewed when `Logiciel` merged into `main`.

The changes I made include:
- Adding the header to all pages (moving `<header/>` up to `App.js`) and adjusting any other UI to fit the header at the top
- Use the user's name from the redux store to display at the top right of the `header` If user is 'logged in' (remember to test this feature you would have to register since log in isn't fully functional yet)
- Added hard-coded teamed and publicationId, this is temporary and only added for us to view what the system looks like when integrated. These values should be removed once proper functionality is implemented.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
`Thunder` had conflicts with `main`


kdav108:
I know this is a `thunder` thing, but since you can see the system integrated as a whole at this stage (before merging into `thunder` and then into `main`) I thought I'll get your opinions as well, Akshay, Richard and Vanessa, and make any changes necessary here instead of creating another branch to do so.

vgoh1 (approved):

rpra0006 (approved):

aksh0001 (approved):
Looks good. Just an innocuous png living inside `client/src/components/layout/user.png` that should be in the `images` directory.

PR #64, opened 2021-04-25
Added Passport JS authentication

AkashSaggar:
# Description 
Implemented Passport JS authentication and more security features.
Upon registration, all passwords will be encrypted before they are stored in the database.
This means that all registered users prior to this point will no longer work, as they were stored without encrypted passwords.

Steps:
- User registers and their password is encrypted
- User logs in
- User can now access protected routes, such as localhost:5000/users/current (this route will give an error if the user is not logged in. Otherwise will return the users details).
- This functionality can be expanded. Any route that needs to be protected just needs to include the authFunctions.ensureAuthenticated middleware

# Type of change 
- [ ] Bug fix
- [ X ] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
There was no proper login system at the moment.

# Solution description
When the user logs in, a cookie will be stored on their machine to identify their session.
So they will not need to log in again until this cookie expires, or we restart the server.
Note: for me I have to change usingRP to false in the /client/src/api/api.js file for this to work.

# Tests 
All unit tests pass: yes
Build successfully: yes
Log in using a new test user and check the console for a success message.
Here is a test user that currently works:
test@test.test
test

# Risk
Identity the risk level(low, medium, high) and indicate which component(s) would be affected if bugs exist 
Medium. The Login page and Register pages would be affected.

### DIFF
--- a/api/src/config/passport.js
+++ b/api/src/config/passport.js
@@ -1,0 +1,29 @@
+const User = require("../models/user.model");
+const bcrypt = require("bcryptjs");
+const localStrategy = require("passport-local").Strategy;
+
+module.exports = function (passport) {
+    passport.use(new localStrategy({usernameField: 'email'}, (email, password, done) => {
+        User.findOne({email: email}).then(user => {
+            if (!user) return done(null, false, { message: 'That email is not registered' });
+
+            bcrypt.compare(password, user.password, (err, result) => {
+                if (err) throw err;
+                if (result === false) {
+                    return done(null, false, { message: 'Password incorrect' });
+                }
+                return done(null, user);
+            });
+        })
+    }));
+
+    passport.serializeUser((user, done) => {
+        done(null, user.id);
+    });
+
+    passport.deserializeUser((id, done) => {
+        User.findById(id, (err, user) => {
+            done(err, user);
+        });
+    });
+};
--- a/api/src/config/passport.js
+++ b/api/src/config/passport.js
@@ -1,8 +1,13 @@
+/**
+ * This module exports a function used by Passport JS for authentication.
+ */
 const User = require("../models/user.model");
 const bcrypt = require("bcryptjs");
 const localStrategy = require("passport-local").Strategy;
 
 module.exports = function (passport) {
+    // Tell Passport to use a localStrategy for authentication (locally saved username/password)
+    // done function is used by Passport to store the found user in the req.user variable
     passport.use(new localStrategy({usernameField: 'email'}, (email, password, done) => {
         User.findOne({email: email}).then(user => {
             if (!user) return done(null, false, { message: 'That email is not registered' });
--- a/api/src/controllers/users.js
+++ b/api/src/controllers/users.js
@@ -14,9 +14,9 @@ const passport = require('passport');
  * @returns: JSON objects of users
  */
 async function getUsers(req, res) {
-  User.find()
-    .then((users) => res.json(users))
-    .catch((err) => res.status(500).json("Error: " + err));
+	User.find()
+		.then((users) => res.json(users))
+		.catch((err) => res.status(500).json("Error: " + err));
 }
 
 /**
@@ -29,18 +29,18 @@ async function getUsers(req, res) {
  * @returns 200: the specified user was found
  */
  async function getUser(req, res) {
-  const {id: _id} = req.params;
+	const {id: _id} = req.params;
 
-  if (!mongoose.Types.ObjectId.isValid(_id))
-      return res.status(400).send('Error: The provided user id is not in a valid hexadecimal format.');
+	if (!mongoose.Types.ObjectId.isValid(_id))
+		return res.status(400).send('Error: The provided user id is not in a valid hexadecimal format.');
   
-  const targetUser = await User.findById(_id);
+	const targetUser = await User.findById(_id);
 
-  if (targetUser == null) { 
-      res.status(404).send('Error: No user matching the id was found.');
-  } else {
-      res.status(200).json(targetUser);
-  }
+	if (targetUser == null) { 
+		res.status(404).send('Error: No user matching the id was found.');
+	} else {
+		res.status(200).json(targetUser);
+	}
 }
 
 /**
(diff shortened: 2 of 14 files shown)


kdav108 (dismissed):
Looks Really good! Would have taken some time to understand and implement passportJS, so great work with that!

I pushed one change to this branch as well so that our app changes page from login to dashboard just so that we can see the effect without looking at the console. 

I just a few suggestions mostly related to our front-end and back-end will work around this branch's changes. Feel free to reply to the comments with your own thoughts.

And yeah, we have to change `usingRp` to false to make it work on our local host, but Scholar is looking into making it a better/smoother process.

api/src/config/passport.js
```diff
@@ -0,0 +1,29 @@
+const User = require("../models/user.model");
+const bcrypt = require("bcryptjs");
+const localStrategy = require("passport-local").Strategy;
+
+module.exports = function (passport) {
+    passport.use(new localStrategy({usernameField: 'email'}, (email, password, done) => {
```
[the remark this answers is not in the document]
AkashSaggar: @kdav108 I've added some comments to explain this.

api/src/controllers/users.js
```diff
@@ -50,17 +51,31 @@ async function getUsers(req, res) {
...
-    password: req.body.password,
-  });
+  User.findOne({ email: req.body.email }, async (err, user) => {
+      if (err) throw err;
+      if (user) {
+        res.status(400).json({message: "Email already registered"});
+        return;
+      }
+      if (req.body.password != req.body.confirmPassword) {
+        res.status(400).json({message: "Passwords do not match"});
```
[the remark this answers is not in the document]
AkashSaggar: @kdav108 I looked online and none of the other status codes are really suitable.
You can see the below link and scroll to the 400 codes
https://kinsta.com/blog/http-status-codes/#a-complete-guide-and-list-of-http-status-codes

401 means Unauthorized. That would be used if the user is trying to access a page where they should be logged in, but they currently aren't. There isn't really a built in status code for "passwords do not match".
I'm thinking we will just have to use `message`, or maybe return another attribute in the JSON with our own "status codes" or something. But maybe that can be done later? As this PR should get merged.

api/src/controllers/users.js
```diff
@@ -50,17 +51,31 @@ async function getUsers(req, res) {
...
  */
 async function addUser(req, res) {
-  const newUser = new User({
-    givenName: req.body.givenName,
-    familyName: req.body.familyName,
-    email: req.body.email,
-    password: req.body.password,
-  });
+  User.findOne({ email: req.body.email }, async (err, user) => {
+      if (err) throw err;
```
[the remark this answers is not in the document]
AkashSaggar: @kdav108 This is done in the users.js
Can't be done in passport.js as that file doesn't have access to `res`

api/src/controllers/users.js
```diff
@@ -50,17 +51,31 @@ async function getUsers(req, res) {
...
+      if (req.body.password != req.body.confirmPassword) {
+        res.status(400).json({message: "Passwords do not match"});
+        return;
+      }
 
-  newUser
-    .save()
-    .then(() => res.status(201).json("User added!"))
-    .catch((err) => res.status(400).json("Error: " + err));
+      const hashedPassword = await bcrypt.hash(req.body.password, 10);
```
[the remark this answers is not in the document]
AkashSaggar: @kdav108 This is the number of salt rounds it should use (amount of processing power).
The higher the number, the more random/secure the salt?
10 is just a reccomended value to use.

api/src/controllers/users.js
```diff
@@ -70,18 +85,31 @@ async function addUser(req, res) {
...
+async function loginUser(req, res, next) {
+  passport.authenticate("local", (err, user, info) => {
+    if (err) throw err;
+    if (!user) {
+      res.status(400).json({message: "User is not registered"});
+      return;
+    }
+
+    req.logIn(user, (err) => {
+        if (err) throw err;
```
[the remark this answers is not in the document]
AkashSaggar: @kdav108 This is done, it now returns status 500

client/package.json
```diff
@@ -17,9 +18,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^4.0.3",
     "redux": "^4.0.5",
-    "redux-thunk": "^2.3.0",
-    "bootstrap": "^4.6.0",
-    "react-bootstrap": "^1.5.2"
```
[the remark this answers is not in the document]
AkashSaggar: @kdav108 Not sure how they got removed but it should be fine now (added back).

client/src/actions/users.js
```diff
@@ -29,3 +29,34 @@ export const addUserAction = (userData) => async (dispatch) => {
     console.error(`Error in posting user data to the api: ${err}`);
```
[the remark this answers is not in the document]
AkashSaggar: @kdav108 This is done

api/src/controllers/users.js
```diff
@@ -92,21 +133,43 @@ async function loginUser(req, res) {
...
+		"_id": req.user["_id"],
+		givenName: req.user.givenName,
+		familyName: req.user.familyName,
+		email: req.user.email
+		}
+
+		res.json(user);
+	} else {
+		res.status(400).json({message: "User not logged in"});
+	}
```
kdav108, line 172: We could probably remove this function entirely, since once the user is logged in, we store their details in the redux global store.

api/src/routes/users.js
```diff
@@ -2,17 +2,21 @@
...
  */
 const usersRouter = require('express').Router();
-
 const usersController = require('../controllers/users');
+const authFunctions = require('../middlewares/authFunctions');
 
 usersRouter.get('/', usersController.getUsers);
 
+usersRouter.get('/current', authFunctions.ensureAuthenticated, usersController.getCurrentUser);
+
```
kdav108, line 11: Related to the previous comment, this won't be needed if we decide to remove this functionality.

client/src/actions/users.js
```diff
@@ -23,9 +23,44 @@ export const addUserAction = (userData) => async (dispatch) => {
...
     // Store user related data in our redux global store
     dispatch({ type: ADD_USER_DATA, payload: data });
+
+    responseCallback(message);
   } catch (err) {
     console.error(`Error in posting user data to the api: ${err}`);
   }
 };
+
+export const getCurrentUserAction = () => async (dispatch) => {
```
kdav108, line 58: Same thing here, if remove `getCurrentUser()` we can remove this too :)

kdav108 (commented):
Nicely done, just have one comment about a function and whether we can remove it, apart from that once the branch is updated with the new changes in thunder we should be good to go!

api/src/controllers/users.js
AkashSaggar: @kdav108 the main point of this function was to act as a demonstration of how the authentication works.
Eg, if a user is not logged in, this function would return an error.
So I think its worth keeping as an example, so people can see how to use the authFunctions.ensureAuthenticated thing.

api/src/routes/users.js
AkashSaggar: Same as above, I think this is a good example for people to see. As no one else would know how to use authFunctions yet.

api/src/controllers/users.js
```diff
@@ -50,17 +51,36 @@ async function getUsers(req, res) {
...
+		}
 
-  newUser
-    .save()
-    .then(() => res.status(201).json("User added!"))
-    .catch((err) => res.status(400).json("Error: " + err));
+		const hashedPassword = await bcrypt.hash(req.body.password, 10);
+
+		const newUser = new User({
+			givenName: req.body.givenName,
```
ssha0054, line 73: `givenName` and `familyName` fields have now been changed to `teamName` and `orgName` in the User model.

api/src/controllers/users.js
```diff
@@ -92,21 +133,43 @@ async function loginUser(req, res) {
...
+ *
+ * @param req request object
+ * @param res response object
+ * @sends: The user object of the currently logged in user
+ */
+async function getCurrentUser(req, res) {
+	if (req.user) {
+		const user = {
+		"_id": req.user["_id"],
+		givenName: req.user.givenName,
```
ssha0054, line 164: Need to update the fields here as well

api/src/controllers/users.js
AkashSaggar: This has been updated

api/src/controllers/users.js
AkashSaggar: This has been updated

kdav108 (approved):
Not sure if we need the `getCurrentUser` method as an example, because there's similar methods already coded which are similar and can act as templates. But I'm approving  cos it's not a big deal tbh.

api/src/controllers/users.js
kdav108: Ahhh I see, that makes sense. We'll keep it then

PR #68, opened 2021-04-30
Backend Twitter Integration

vgoh1:
# Description 
Stores a twitter handle with the team doc so that it can be retrieved by the frontend to display tweets associated with the account. Also implemented a get method for team doc for the time being.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
We want to be able to show tweets of a twitter account on the frontend, I found that an [embedded timeline](https://developer.twitter.com/en/docs/twitter-for-websites/timelines/overview) exists which means we don't have to implement our own components for it, and just need to retrieve the twitter handle for the team to show tweets.

# Solution description
It validates whether the given twitter handle is valid is or not by hitting one of the Twitter endpoints that searches for users by username. It also validates whether the team id being passed in exists or not. 

Also note that there is duplicated code to validate the team id, wasn't sure how to refactor it in a way that made sense because I needed it to return early if the team id wasn't valid. If anyone knows how to refactor it the right way, let me know!

# Tests 
Example curl when hitting the store handle endpoint (keep in mind you'll need a bearer token from the Twitter API in your .env file): curl --location --request PATCH 'http://localhost:5000/team/608a0d11ef93efa524bb2781/twitter-handle' \
--header 'Content-Type: application/json' \
--data-raw '{
    "twitterHandle": "JamesTurner_42"
}'

Example curl for hitting get team: curl --location --request GET 'http://localhost:5000/team/606bb5c022201f529db920ca'

Valid twitter handle
image

Invalid twitter handle
image

Removing twitter handle
image

Team doesn't exist
image

Invalid teamid format
image

N/A

# Risk
Low - endpoints are new so they shouldn't impact on anything right now


api/src/controllers/team.js
```diff
@@ -0,0 +1,85 @@
...
+ * Associates a twitter handle with a team on the /team/twitter-handle/:team-id endpoint. 
+ * @param {*} req request object, containing the team_id in the url and twitter handle in the body
+ * @param {*} res response object
+ * @returns 200: successful added twitter handle to team
+ * @returns 400: team id is not in a valid hexadecimal format
+ * @returns 404: team is not found, or handle is invalid
+ * @returns 500: error trying to update the document in db
+ */
+async function storeHandle(req, res) {
+    const {team_id: team_id} = req.params;
```
aksh0001, line 25: nitpick: `{team_id: team_id}` redundant renaming
see: https://www.barrymichaeldoyle.com/destructuring/

api/src/controllers/team.js
```diff
@@ -0,0 +1,85 @@
...
+    const {twitterHandle: handle} = req.body;
+
+    // validate if the team exists
+    if (mongoose.Types.ObjectId.isValid(team_id)) {
+        var foundTeam = await Team.findById(team_id);
```
aksh0001, line 30: minor: prefer `let` over `var`
see: https://www.geeksforgeeks.org/difference-between-var-and-let-in-javascript/#:~:text=var%20and%20let%20are%20both,program%20as%20compared%20to%20let.

api/src/routes/team.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * This module defines the endpoints for the "/twitter" route and exports the corresponding Router.
+ */
+
+const teamRouter = require("express").Router();
+
+const teamController = require('../controllers/team');
+
+teamRouter.patch('/twitter-handle/:team_id', teamController.storeHandle);
```
aksh0001, line 9: The route naming seems a bit awkward here, imo. Do you think it's more appropriate to shuffle the order to `/:team_id/twitter-handle`? My line of thinking was that in the route: first identify the id of the team -> then define the operation that the route handles (in this case `twitter-handle`).

api/src/controllers/team.js
```diff
@@ -0,0 +1,85 @@
...
+        if (foundTeam == null) {
+            return res.status(404).send(`Error: No team found with given id.`);
+        }
+    } else {
+        return res.status(400).send("Error: Given team id is not in a valid hexadecimal format.");
+    }
```
aksh0001, line 79: for refactoring this duplicated block, I'm thinking we can use middleware.
the flow I'm thinking of: 
```
request -> validate mongo object id middleware function -> if success, validate team exists middleware function -> if success, `next()` into the corresponding handler.
```
Similar to what we're doing for the publications: 
https://github.com/Researchify/Researchify/blob/main/api/src/routes/publications.js#L14
https://github.com/Researchify/Researchify/blob/main/api/src/middleware/publications.js

Though this may be tedious, and I'm okay with leaving as is as it's just a couple of lines.

api/src/routes/team.js
vgoh1: The reason why I did it like this is cause it doesn't let /team/:team_id for getting the team doc route so I switched it around. If it's /team/:team_id and then /team/:team_id/twitter-handle, when I'm trying to hit the first one it thinks I'm trying to hit the second one

aksh0001 (approved):
Nice work!

api/src/controllers/team.js
```diff
@@ -0,0 +1,65 @@
...
+ * @returns 200: the team was found
+ * @returns 404: team is not found
+ * @returns 400: team id is not in a valid hexadecimal format
+ */
+async function getTeam(req, res) {
+    const {team_id} = req.params;
+    let foundTeam = await Team.findById(team_id);
+
+    return res.status(200).send(foundTeam);
+
```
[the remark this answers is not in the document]
vgoh1: The lines you've highlighted are just getting the team document from the db - so technically unrelated for the twitter integration but I just added it so it'd be easier to test on the front end. If the handle was invalid, it wouldn't be added to the doc. Unless you're talking about the token used to connect to the twitter API?

api/src/controllers/team.js
```diff
@@ -0,0 +1,65 @@
...
+    const {team_id} = req.params;
+    const {twitterHandle: handle} = req.body;
+    let foundTeam = await Team.findById(team_id);
+
+    if (handle.length == 0) {  // remove the handle from the doc
+        foundTeam.twitterHandle = ""
+        console.log(foundTeam);
+    } else {  // update the handle
+        // validate the handle by getting user id
+        let response = await axios.get("https://api.twitter.com/2/users/by/username/" + handle, options)
```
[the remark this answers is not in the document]
vgoh1: Ah yup, I didn't realize that there was a limit for the twitter handle - will add

pmyip3 (approved):

PR #78, opened 2021-05-12
[Logiciel into Main] Sidebar added, profile management page editted

jfan0010:
# Description 
- Added sidebar
- Modified profile management page
- Deleted unwanted files and images
- Changed background colour of header, added border for it
- Removed header/sidebar from homepage/login/signup
- Removed links in header, added links in sidebar
- A bit of refactoring

See screenshots below.

# Type of change 
- [x] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
-

# Solution description
-

# Tests 
All unit tests pass?
Yes

Builds successfully?
Yes

Header and Sidebar
image

Profile Management Page:
image

Homepage without header and sidebar:
image

# Risk
Might occur faulty alignment when changing screen size.


client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -61,25 +50,57 @@ function ProfileInfoEdit () {
                              <Form.Control type="text" placeholder="+62123456789" />
                         </Form.Group>
 
+                        <Form.Group controlId="formWebUrl">
+                             <Form.Label>Website URL</Form.Label>
+                             <InputGroup className="mb-3">
+                                <Form.Control 
+                                    type="text"
+                                    placeholder="Team custom URL"
+                                    aria-label="Team custom URL"
```
kdav108, line 59: Is this for their own website that they build using Researchify? If so, should we place this outside of Profile page and somewhere in a separate page dedicated for website settings?

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -61,25 +50,57 @@ function ProfileInfoEdit () {
...
+                        <div className= "my-1">
+                            <Button 
+                                color="primary"
+                                className= "mr-2"
+                                onClick={(event) => {
+                                    event.preventDefault();
+                                    alert('You have updated your profile')
+                                    /**
+                                     * Update user information here onclick button
+                                     */
```
kdav108, line 77: I know this isn't complete yet, but just putting it out there when we do implement this, if we're thinking of keeping this `alert`, I think we could instead use something like a `Toast` instead. Here's a really simple, beautiful and my favourite Toast library that I've introduce in `thunder`: https://react-hot-toast.com, feel free to have a look :)

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -26,21 +26,10 @@ function ProfileInfoEdit () {
...
-                                <Form.Group controlId="formUserLastName">
-                                    <Form.Label>Last Name</Form.Label>
-                                    <Form.Control type="text" placeholder="Doe" />
-                                </Form.Group>
-                            </Col>
-                        </Row>
+                        <Form.Group controlId="formUserFirstName">
+                            <Form.Label>Research Group Name</Form.Label>
+                            <Form.Control type="text" placeholder="Team name.." />
+                        </Form.Group>
```
kdav108, line 32: `Thunder` is thinking about having two separate information, a Team name, as we have here and an Organisation Name. So for example, a team name maybe something the Research group chose, like DeepHackers, and the Organisation might be Monash or a corporate company. So could we add another field here to change the team's Organisation name?

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -26,21 +26,10 @@ function ProfileInfoEdit () {
...
-                            
-                            <Col>
-                                <Form.Group controlId="formUserLastName">
-                                    <Form.Label>Last Name</Form.Label>
-                                    <Form.Control type="text" placeholder="Doe" />
-                                </Form.Group>
-                            </Col>
-                        </Row>
+                        <Form.Group controlId="formUserFirstName">
+                            <Form.Label>Research Group Name</Form.Label>
```
kdav108, line 30: `Thunder` is thinking about having two separate information, a Team name, as we have here and an Organisation Name. So for example, a team name maybe something the Research group chose, like DeepHackers, and the Organisation might be Monash or a corporate company. So could we add another field here to change the team's Organisation name?

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -26,21 +26,10 @@ function ProfileInfoEdit () {
...
+                            <Form.Control type="text" placeholder="Team name.." />
```
kdav108, line 31: I know this isn't really related to this PR, but if you're going to make changes to it regarding other changes requested, could we add the current value in place of the placeholders? So using something like `const teamName = useSelector(state => state.user?.teamName)` we can get the team name from the Redux store and then in place of the `Team name...` place holder we can probably use `${teamName}`. And this relates to the country and phone number as well.

kdav108 (changes requested):
Nice to see our chosen colours on the website! It looks really good imo 😍. Great job guys!
 
I was wondering what colour do we use for highlighting the page in the Navigation bar, it seems to go well with our header colour. Maybe we can make that our secondary colour for buttons. Thoughts @frontend people?

Also, I was wondering if we can make the navigation bar closable, so it's not always taking up our window space.

Apart from these I just have a few requests regarding the Team profile page.

client/package.json
```diff
@@ -3,11 +3,14 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@material-ui/core": "^4.11.4",
+    "@material-ui/icons": "^4.11.2",
```
aksh0001, line 7: I'm confused: are we using bootstrap instead of material-ui? if so, do we need these dependencies still?

client/src/components/dashboard/Dashboard.js
```diff
@@ -1,24 +1,37 @@
 import Container from "react-bootstrap/Container"
 
-import PageCard from "./PageCard"
-import Layout from '../layout/Layout'
 import './Dashboard.css'
 
 
-function PageContent() {
+function Dashboard() {
```
aksh0001, line 6: prefer arrow functions to be consistent

client/src/components/layout/Sidebar.js
```diff
@@ -0,0 +1,36 @@
+import { Navbar, Nav, Image } from "react-bootstrap"
```
aksh0001, line 1: All these imports are unused according to the static analyzer, can remove if not needed

aksh0001 (changes requested):
We should remove imports that are not being used (+ dependencies) and ensure consistent style w.r.t. writing components. Since this is a merge to the mainline, I think we can make more changes before merging.

client/src/components/dashboard/Dashboard.js
kdav108: Do you mean using `const Dashboard = () => {` instead? I personally find using `function Deashboard() {` easier to understand since it's consistent with other languages and we don't really need it be a variable function in this particular case. Though I'm not an expert in JS and maybe doing it with arrow functions is better??

riordanalfredo: Yes, if this is a component, React best practices recommend using the arrow function. Besides, it will support React Hook if you need them later. https://frontarm.com/james-k-nelson/when-to-use-arrow-functions/

jfan0010:
@kdav108 The highlight colour in the navigation bar is  #2B61F2,
and for the navbar, it is relatively easier to build a dropdown menu from the header on top, but I doubt that if we can easilybuild a closable drawer from side using only react-bootstrap, if we're to implement a drawer, might need to use some third party library?

kdav108:
@jfan0010, I found the following article which shows how we can close and open the navigation bar. I had a quick scan and it seems like a doable option. Have a look and we'll see if it's easily implementable or not:
https://medium.com/swlh/how-to-make-a-side-navigation-bar-in-reactjs-c90747f3410c

client/src/components/dashboard/Dashboard.js
kdav108: Damn I actually never knew this lol. Although on that second point, I believe you can use React Hooks within non-arrow functions as well. 
Also every major function in `thunder` branch have been built using `function()` and not `() => {}`. Is this something important that we should go back and refactor those things, or we just move forward with arrow functions from now on. I think the latter is more wise since we still have a lot of features to go.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: Field deleted, this can be add outside profile settings

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: Changed alert to toast, thanks for suggestion

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: Added new fields

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: Followed suggestion to use Redux to fetch value from state to fill in placeholders.

client/package.json
rpra0006: Yes we are, unused packages have been deleted.

client/src/components/dashboard/Dashboard.js
rpra0006: @Researchify/logiciel has changed all React files associated with our commits to arrow functions.

rpra0006:
> @jfan0010, I found the following article which shows how we can close and open the navigation bar. I had a quick scan and it seems like a doable option. Have a look and we'll see if it's easily implementable or not:
> https://medium.com/swlh/how-to-make-a-side-navigation-bar-in-reactjs-c90747f3410c

@kdav108 We will discuss again with front-end people whether to have a fixed sidebar, or drawer.
image
In Wordpress, they use fixed sidebars.

client/src/components/dashboard/Dashboard.js
aksh0001: Can you assign one of your teammates to do it @kdav108 as I think it should be a very simply refactor that shouldn't take too long? I personally think one thing we can do well is to maintain a clean code base at all times to avoid technical debt piling up and my bad for not picking up on not using the arrow functions before. I must have missed it when doing my review for that PR.

client/src/components/App.js
```diff
@@ -11,24 +11,43 @@ import ProfileInfoEdit from './profileInfoEdit/ProfileInfoEdit';
...
 import Login from './auth/Login';
-import Header from './layout/Header'
+import Header from './layout/Header';
+import Sidebar from './layout/Sidebar';
+import { Container, Col, Row } from 'react-bootstrap';
+
+import './layout/Layout.css'
 
 import PublicationPage from './publications/PublicationPage';
+import { Fragment } from 'react';
```
aksh0001, line 21: Isn't `Fragment` the same as `<></>`? If so, prefer using the latter to maintain consistency.

client/src/components/App.js
```diff
@@ -11,24 +11,43 @@ import ProfileInfoEdit from './profileInfoEdit/ProfileInfoEdit';
...
+                            <Sidebar />
+                        </Col>
+                        <Col className="page-content-wrapper" md={10} lg={10} xl={11}>
+                            <Route path="/publications/team/:teamId" exact component={PublicationPage}/>
+                            <Route path={urls.dashboard} exact component={Dashboard} />
+                            <Route path="/dashboard/profile" exact component={ProfileInfoEdit} />
+                        </Col>
+                        </Row>
+                    </Container>
+                </Fragment>
```
aksh0001, line 49: Is there a way we can refactor this block to a separate component? How about encapsulating it all in `Header` itself?
I think we should keep the root `App` component clean and avoid defining UI here unless we don't have a choice.

client/src/components/App.js
```diff
@@ -11,24 +11,43 @@ import ProfileInfoEdit from './profileInfoEdit/ProfileInfoEdit';
...
+import './layout/Layout.css'
 
 import PublicationPage from './publications/PublicationPage';
+import { Fragment } from 'react';
 
 const App = () => {
+    const urls={
+        dashboard: "/dashboard",
+        profile: "/dashboard/profile"
+    }
```
aksh0001, line 27: minor: this object doesn't seem to be adding much readability to the route paths. I'd say we can scrap it entirely or store each route path into this object as a constant.

client/src/components/layout/Header.js
```diff
@@ -2,34 +2,33 @@ import React from "react"
 import { Navbar, Nav, Image } from "react-bootstrap"
```
aksh0001, line 2: Unused `Image` (SA - static analyzer warning)

client/src/components/layout/Header.js
```diff
@@ -2,34 +2,33 @@ import React from "react"
...
 import { Link } from 'react-router-dom'
 import {useSelector} from 'react-redux';
+import {BsPeopleCircle} from 'react-icons/bs'
 
 import "./Header.css"
 
-import profilePicture from "../../images/user.png"
+import { Fragment } from "react";
```
aksh0001, line 9: Same as above, `<></>`

client/src/components/layout/Sidebar.js
```diff
@@ -0,0 +1,38 @@
+import "./Sidebar.css"
```
aksh0001, line 1: File headers would be nice to describe what this component is.

client/src/components/dashboard/Dashboard.js
```diff
@@ -1,24 +1,16 @@
 import Container from "react-bootstrap/Container"
```
aksh0001, line 1: File headers would be nice

client/src/components/layout/Header.js
aksh0001: Also, missing file headers.

client/src/components/layout/SidebarData.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
```
aksh0001, line 1: File headers.

client/src/components/layout/SidebarData.js
```diff
@@ -0,0 +1,39 @@
...
+import {BsFillGridFill, BsFillHouseDoorFill, BsBookHalf, 
+    BsPerson, BsGearFill, BsFillQuestionCircleFill} from "react-icons/bs";
+
+/*
+    List of navigations inside the sidebar. Change the link to Nav.Link either here or in Sidebar.js.
+*/
+export const SidebarData = [
```
aksh0001, line 8: Can we `export default` this since this is the only thing exported?

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -5,86 +5,98 @@
```
aksh0001, line 5: File headers.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -5,86 +5,98 @@
...
+const profileUpdated = () => toast.success('Profile has been successfully updated')
+const profileDeleted = () => toast.error('Profile has not been deleted')
 
+const ProfileInfoEdit = () => {
+
+    const groupName = useSelector(state => state.user?.groupName); 
+    const orgName = useSelector(state => state.user?.orgName); 
+    const country = useSelector(state => state.user?.country); 
+    const email = useSelector(state => state.user?.email); 
+    const phoneNum = useSelector(state => state.user?.phoneNum);
```
aksh0001, line 24: Is it better to combine this by just selecting the `user` data from the redux store as opposed to the individual attributes for a user? Or is there a reason to do it this way?
i.e. `const user = useSelector(state => state.user);`

aksh0001 (approved):
LGTM - I don't want to block you guys as the comments are to do with style/conventions and we are admittedly short on time - but we should set aside some time to address these.

client/src/components/App.js
jfan0010: I found that the difference between `Fragment` and `<></>` is that `Fragment` allows key to be passed to `Fragment`, and `<></>` does not support keys or attributes. Not sure if we gonna need to pass attributes in the component later, but should we use `Fragment` in our code for reusability of components? 

Ref: [React - Short Syntax and Keyed Fragments](https://reactjs.org/docs/fragments.html#:~:text=table%3E-,Short%20Syntax,Live,-Demo)

client/src/components/App.js
vgoh1: +1

jfan0010: Will try to refactor this part in next sprint.

client/src/components/layout/Header.js
```diff
@@ -2,34 +2,33 @@ import React from "react"
...
-            </Nav>
-        </Navbar>
+        <Fragment>
+            <Navbar className="header" sticky="top">
+                <Navbar.Brand><Link className="header-brand" to={urls.dashboard}>{title}</Link></Navbar.Brand>
+                <Nav className="mr-auto" />
+                <Nav>
+                    <Nav.Link className="header-profile">
+                        <Link className="header-link" to="/dashboard/profile">
+                            <BsPeopleCircle className="header-profile-icon" /> {userName}
```
vgoh1, line 26: Just wondering, is the username supposed to show next to the profile icon in the top right? It might be good to hardcode a name for demo purposes

vgoh1 (approved):

vgoh1:
Probably not a big issue at the moment and can be fixed later (either by scholar or logiciel probs) but the publications page clips against the sidebar for some reason if you scroll right, not sure if its because of the twitter box. Ideally you shouldn't be able to scroll right FYI @aksh0001 
image

client/src/components/layout/Header.js
jfan0010: How about the profile editing page? Do we also put a hardcoded value in profile edit page for now? XD

vgoh1: Yeah, I suppose so for demo purposes, like pretend we have a default user logged in

kdav108 (approved):
Looks good to me Logiciels!

PR #82, opened 2021-05-14
Modified user model to include 'teamName' and 'orgName'

ssha0054:
# Description 
A summary of the change 
Modified the user model so that it now has fields for team name and organisation. Removed the previous fields for given name and family name as users no longer have their own account. 
Note: All the existing users in our database have to be removed or modified manually as they have the `givenName` and `familyName` fields present over `teamName` and `orgName`. 

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
Users from the same team now use a team account to log in to Researchify. Previously we assumed each user will have their own account so we had `givenName` and `familyName` fields in our model.

# Solution description
Removed `givenName` and `familyName` fields and replaced them with `teamName` and `orgName`. 

# Tests 
All unit tests pass?
N/A
Builds successfully?
Yes

Tested all the api requests we had for the user model on insomnia and they all work fine. Can attach screenshots if requested.

# Risk
Low

### DIFF
--- a/api/src/controllers/users.js
+++ b/api/src/controllers/users.js
@@ -54,8 +54,8 @@ async function getUsers(req, res) {
  */
 async function addUser(req, res) {
   const newUser = new User({
-    givenName: req.body.givenName,
-    familyName: req.body.familyName,
+    teamName: req.body.teamName,
+    orgName: req.body.orgName,
     email: req.body.email,
     password: req.body.password,
   });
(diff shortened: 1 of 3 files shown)


kdav108 (approved):
Looks good to me! Just wondering do we not need fake team name and org name in the seederScript anymore?

ssha0054:
> Looks good to me! Just wondering do we not need fake team name and org name in the seederScript anymore?

The seederScript will just populate the table with data every time we run the application, assuming that data does not already exist in the database. We can just manually add in a fake user instead so this is not an issue.

PR #83, opened 2021-05-14
Get publications using scholarly

vgoh1:
# Description 
Using the scholarly python library, get the publications associated with an author given their google scholar user id

# Type of change 
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
Before we had to scrape the results using a node package but that wasn't ideal so we decided to do it in python instead where scholarly does the scraping for us.

# Solution description
AWS Lambda has issues with the lxml lib that scholarly is dependent on so we had to do it via a container image. The base container image is taken from the base [AWS Lambda python image](https://hub.docker.com/r/amazon/aws-lambda-python), and then it installs scholarly and copies over the function code. It needs to edit the scholarly source code because of this error:
https://github.com/hellysmile/fake-useragent/issues/81
https://stackoverflow.com/questions/57531742/getting-errors-when-using-google-search-api-on-python

# Tests 
In the directory - make sure you have Docker installed:
To build the docker image: `docker -t build test-image .`
Then run a container with a port: `docker run -p 9000:8080 test-image`
Then run in a new terminal: `curl -XPOST "http://localhost:9000/2015-03-31/functions/function/invocations" -d '{"author_id":"eRbvWqYAAAAJ"}'`

Results:
The docker terminal:
`START RequestId: 0a9494ca-8601-4bb5-b009-b420b5392ffa Version: $LATEST
eRbvWqYAAAAJ
END RequestId: 0a9494ca-8601-4bb5-b009-b420b5392ffa
REPORT RequestId: 0a9494ca-8601-4bb5-b009-b420b5392ffa  Duration: 24380.50 ms   Billed Duration: 24400 ms       Memory Size: 3008 MB    Max Memory Used: 3008 MB`
The other terminal should output something like below:
image

N/A

# Risk
LOW, standalone feature


importpub/gscholar.py
```diff
@@ -0,0 +1,20 @@
+from scholarly import scholarly
```
aksh0001, line 1: File headers would be nice.

importpub/gscholar.py
```diff
@@ -0,0 +1,20 @@
...
+import json
+
+def getPublications(author_id):
+    """
+    This function uses the scholarly library to retrieve publications associated with an author.
+    :param author_id: google scholar user id extracted from their profile url
+    :return: json string of publications in a list
+    """
+    search_query = scholarly.search_author_id(author_id)
+    authorObject = scholarly.fill(search_query, publication_limit=10)
```
aksh0001, line 11: camelCase naming used here.

importpub/gscholar.py
```diff
@@ -0,0 +1,20 @@
...
+    pub_list = []
+
+    for i in range(len(authorObject["publications"])):
```
aksh0001, line 14: nitpick: probably don't need a `range` loop here. Instead:
`for pub in author_object['publications']:`

importpub/handler.py
```diff
@@ -0,0 +1,13 @@
+from gscholar import getPublications 
+
+def lambda_handler(event,context):
+    """
+    This function is called when the lambda is invoked.
+    :param event: a json object containing the author_id
+    :param context: object with context properties see: https://docs.aws.amazon.com/lambda/latest/dg/python-context.html
+    """
+    author_id = event["author_id"]
+    print(author_id)
```
aksh0001, line 10: minor: I think we can use logging instead of print statements: https://docs.python.org/3/howto/logging.html

importpub/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM public.ecr.aws/lambda/python:3.8
+
+RUN pip3 install scholarly
+
+# Copy function code
+COPY handler.py ./
+COPY gscholar.py ./
+
+RUN cd /var/lang/lib/python3.8/site-packages/scholarly && \
+    sed -i '370s/()/(use_cache_server = False, verify_ssl=False)/' _proxy_generator.py
```
aksh0001, line 10: Does this correctly edit the library contents or did you end up going with editing the file in the running container and committing it as a new image?

aksh0001 (approved):
Looks good to me!

importpub/Dockerfile
vgoh1: This does edit the library contents without me having to go into it to change it and make a new image

PR #91, opened 2021-05-18
Pmyip/improvements for publications page

pmyip3:
# Description 
Some improvement features implemented on the publication page

# Type of change 
- [ ] Bug fix
- [X] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
The current publication page have some potential issues and had been raised by team members 

# Solution description
1. Replace the add and import publication logo to buttons with text
2. Change the individual publication header to green when a new publication is added
     - it will last for 2.5 seconds currently
3. Displaying a spinning icon when fetching publication from server
4. Showing a message when there is no publication for a team
    * to test this feature, need to change the teamId in teamReducer to "609f5ad827b1d48257c321d5"

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

Change the individual publication header to green when a new publication is added
image

Displaying a spinning icon when fetching publication from server
image

Showing a message when there is no publication for a team
image

N/A

# Risk
Low.


rsun0013 (approved):

vgoh1:
2 things:
- When I run from your branch the add and import buttons are on top of each other rather than next to each other, did you forget to push maybe? 
- If I pass in a teamid that doesn't exist in the teamReducer.js, the loading goes on forever despite the backend returning 404 not found. Could we handle this in case it does happen?

pmyip3:
> 2 things:
> 
> * When I run from your branch the add and import buttons are on top of each other rather than next to each other in your screenshot, did you forget to push maybe?
- I think it's beacuse my monitor is wider than normal, so I don't get the same issue as yours.  I'll fix that. 

> * If I pass in a teamid that doesn't exist in the teamReducer.js, the loading goes on forever despite the backend returning 404 not found. Could we handle this in case it does happen?
- you mean passing a teamid does not exist in database? Good point. I am assuming that a user will always be assigned to a team.  So when a user sign in succesfully, the corresponding teamId will be sent from the database and store in the reducer. So we can assume that the teamId always exists in the database ( the hardcoded the teamId in reducer just assuming the user has signed in and received the teamId from database ).  Unless a user does not have to be assigned to a team? Will it be the case? Then I'll need to display a warning message here.

vgoh1:
> * you mean passing a teamid does not exist in database? Good point. I am assuming that a user will always be assigned to a team.  So when a user sign in succesfully, the corresponding teamId will be sent from the database and store in the reducer. So we can assume that the teamId always exists in the database ( the hardcoded the teamId in reducer just assuming the user has signed in and received the teamId from database ).  Unless a user does not have to be assigned to a team? Will it be the case? Then I'll need to display a warning message here.

I think assuming the teamid always exists in the database is okay because I actually can't think of a situation where a bad teamid will be passed to the frontend and it was just me trying to break your code. It's probably some specific business logic that isn't too important right now, so don't worry about it then.

vgoh1 (approved):
Looks good to me!

PR #94, opened 2021-05-20
Update Registration Form

kdav108:
# Description 
Update the Registration Form to take team and organisation name instead of family and given name.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
The client requirements change led to our existing Registration Form being out-of-date. We now don't need family and given name, since the registration is for a team now.

# Solution description
Replace the appropriate text input fields.

# Tests 
Yes

### DIFF
--- a/client/src/components/App.js
+++ b/client/src/components/App.js
@@ -1,36 +1,43 @@
 /**
  * Root component.
  */
-import React from 'react';
-import {BrowserRouter, Switch, Route} from 'react-router-dom';
+import React from "react";
+import { Toaster } from "react-hot-toast";
+import { BrowserRouter, Switch, Route } from "react-router-dom";
 
-import Auth from './auth/Auth';
-import Home from './home/Home';
-import Dashboard from './dashboard/Dashboard';
-import ProfileInfoEdit from './profileInfoEdit/ProfileInfoEdit';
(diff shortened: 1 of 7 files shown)


client/src/components/App.js
```diff
@@ -1,36 +1,43 @@
 /**
  * Root component.
  */
-import React from 'react';
-import {BrowserRouter, Switch, Route} from 'react-router-dom';
+import React from "react";
+import { Toaster } from "react-hot-toast";
```
ssha0054, line 5: I think you haven't pushed the updated `package.json` file. I got an error saying `Module not found: Can't resolve 'react-hot-toast' in....` when I tried running it. I tried doing `npm install` and it didn't fix it so what I did was `npm install react-hot-toast` and that added a new line in `package.json`. Would ideally be better to add this in before doing the merge.

ssha0054 (commented):
Code looks fine to me however when i tried testing it, I noticed that the `teamName` and `orgName` fields were not being pushed to the database. 
- Below is an example of the inputs I tried testing with.
image
- Image below is what is pushed to the database
image

ssha0054 (approved):
Everything works well now!

PR #99, opened 2021-05-20
Add API Request to update team details 

ssha0054:
# Description 
Added a `PATCH` api request to allow updating a team's details.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
Needed a way to add/update information on the `Team` model (i.e. need to be able to add GitHub Token to team's model at a later stage).

# Solution description
- Added a new field in the `Team` model called `githubToken`, which would store the GitHub token.
- Created a `PATCH` api requests that updates the team details with the provided JSON object. This `PATCH` request is on the endpoint /team/:id , where ':id' is the id of the team. This supports updating all or part of the team's current details and also gives us a way to be able to store a GitHub token in the `Team` model.

# Tests 
All unit tests pass?
Yes

Builds successfully?
Yes

- Below is a `GET` request for a team.
image
- Image below shows us make a `PATCH` request for that team. (Note: We provide a JSON object with the changes we want to make. Here we are trying to update the GitHub Token)
image
- Image below shows a `GET` request (the same request made earlier), and you can see the `githubToken` field is updated.
image

# Risk
Low

### DIFF
--- a/api/src/controllers/team.js
+++ b/api/src/controllers/team.js
@@ -4,7 +4,6 @@
 const axios = require("axios");
 const mongoose = require("mongoose");
 const Team = require("../models/team.model");
-const teamRouter = require("../routes/team");
 
 
 const options = {
--- a/api/src/controllers/team.js
+++ b/api/src/controllers/team.js
@@ -2,6 +2,9 @@
  * This module contains handlers for the "team" route.
  */
 const axios = require("axios");
+const mongoose = require("mongoose");
+const Team = require("../models/team.model");
+const teamRouter = require("../routes/team");
 
 
 const options = {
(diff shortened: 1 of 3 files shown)


api/src/controllers/team.js
```diff
@@ -59,4 +61,28 @@ async function getTeam(req, res) {
...
+async function updateTeam(req, res) {
+    const { id: _id } = req.params;
+    const updates = req.body;
+    const options = {new: true};
+
+    if (!mongoose.Types.ObjectId.isValid(_id))
+    return res.status(404).send("Error: No team with that id.");
+
+    try {
+        const result = await Team.findByIdAndUpdate(_id, updates, options);
```
kdav108, line 81: Is `findByIdAndUpdate` our method or some in-built one?

api/src/models/team.model.js
```diff
@@ -21,6 +21,9 @@
     },
      twitterHandle: {
         type: String
+    },
+     githubToken: {
+        type: String
```
kdav108, line 26: Are you adding the post and get request for this in another PR?

api/src/controllers/team.js
ssha0054: It's an inbuilt method from Mongoose.

api/src/models/team.model.js
ssha0054: This patch request was created for the functionality of posting the GitHub token to a current team. You can just pass in a JSON containing only the githubToken and send a patch request to store the token (refer to images above). There's already a GET request for team/id endpoint which will return a json of all the details for that team so you can use this to get the token.

kdav108: Ahh right, sorry didn't see the screenshot hahaha

kdav108 (approved):
LGTM!

PR #100, opened 2021-05-20
Optimizing scholarly lib

vgoh1:
# Description 
Parallelize getting the publications

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
Currently importing publications takes way too long, it takes 20 seconds to get author + 10 publications. Searching up the author takes around 5 seconds which is unavoidable so we have to optimize getting the publications. This is done by getting the info for each publication in parallel since they're independent.

# Solution description
Using the multiprocessing.Pipe function, it creates a parent and child pipe for each publication to get info for. The child fetches the info and then sends it to the parent. The processes are started in a for loop so they don't exactly start at the same time but close enough. The code doesn't continue until all the processes are finished. The parent receives it and puts the publication data into a list.

# Tests 
You can run it locally using Chunyang's Gscholar id: 3tyGlPsAAAAJ
And it should spit out something like:
image

Findings/times are documented here (I tried with joblib before but it wasn't working in the container/lambda): https://docs.google.com/document/d/1CrZi1HxeSRmjTsOq30PnOb9OFWUmn_djxgjCAYKQZRs/edit
I followed the example given here: https://aws.amazon.com/blogs/compute/parallel-processing-in-python-with-aws-lambda/

# Risk
LOW, standalone

### DIFF
--- a/importpub/gscholar.py
+++ b/importpub/gscholar.py
@@ -2,7 +2,13 @@
 This file contains the functions relating to getting publications from google scholar.
 """
 from scholarly import scholarly
+from joblib import Parallel, delayed
 import json
+
+def fill_pub(pub):
+    filled_pub = scholarly.fill(pub)
+    del filled_pub["source"]
+    pub_list.append(filled_pub)
 
 def get_publications(author_id):
     """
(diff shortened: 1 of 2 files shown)


importpub/gscholar.py
```diff
@@ -14,10 +21,23 @@ def get_publications(author_id):
...
+    parent_connections = []
+
     for pub in author_object["publications"]:
-        current_pub = scholarly.fill(pub)
-        del current_pub["source"]
-        pub_list.append(current_pub)
+        parent_conn, child_conn = Pipe()
+        parent_connections.append(parent_conn)
+        process = Process(target=fill_pub, args=(pub,child_conn))
+        processes.append(process)
```
aksh0001, line 31: nice job using IPC to get data from parent to child via pipes!
Another alternative would be using shared memory: https://docs.python.org/3/library/multiprocessing.shared_memory.html and locks, but this looks good.

aksh0001 (approved):

PR #108, opened 2021-05-23
Update team model

kdav108:
# Description 
Update the team model, so we now store team name and organisation name instead of family and given name.

Update the registration Form page to reflect this.

Add githubToken and githubUsername fields to the team model.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
Client required one login credential, so given and family name are no longer required.
And we also need to store user's GitHub token and username

# Solution description
Add team and org names and GitHub token and username fields.

# Tests 
All unit tests pass?
Builds successfully?
Yes


api/src/controllers/team.js
```diff
@@ -59,4 +61,28 @@ async function getTeam(req, res) {
...
\ No newline at end of file
+/**
+ * Handles a PATCH request to update a team's details on the endpoint /team/:id.
+ * @param {*} req request object - team id in url, json object containing atleast one field from team model.
+ * @param {*} res response object - updated team object
+ * @returns 200: returns updated team details
+ * @returns 404: team not found
+ * @returns 400: error updating team
+ */
+async function updateTeam(req, res) {
```
kdav108, line 72: This Patch request is so we can add GitHub token and username

api/src/controllers/team.js
```diff
@@ -59,4 +61,28 @@ async function getTeam(req, res) {
...
+    const { id: _id } = req.params;
+    const updates = req.body;
+    const options = {new: true};
+
+    if (!mongoose.Types.ObjectId.isValid(_id))
+    return res.status(404).send("Error: No team with that id.");
```
aksh0001, line 78: nit: `return` indentation

aksh0001: also, out of scope for this PR, but I'm curious about the `getTeam` controller function. Is this fully fleshed out?
https://github.com/Researchify/Researchify/blob/main/api/src/controllers/team.js#L56-L60

aksh0001: out of scope for this PR, but since thunder is looking at the teams functionality, is this controller function fleshed out? https://github.com/Researchify/Researchify/blob/main/api/src/controllers/team.js#L56-L60
The skeleton was coded by @vgoh1 as a stopgap and you guys may consider implementing this as you see fit.

api/src/models/user.model.js
```diff
@@ -5,8 +5,8 @@ const mongoose = require('mongoose');
 
 
 const userSchema = new mongoose.Schema({
-    givenName: { type: String, required: false },
-    familyName: { type: String, required: false },
+    teamName: { type: String, required: true },
+    orgName: { type: String, required: false },
```
aksh0001, line 9: What's the difference between `teamName` and `orgName`? We should document this somewhere (e..g https://docs.google.com/document/d/1U68bJV0POrh_wDXem_hZ_PrivoVT6qZCPC8vw7Dg5uQ/edit)

api/src/routes/team.js
```diff
@@ -12,4 +12,6 @@ teamRouter.patch('/:team_id/twitter-handle', teamMiddleware.validateTeamId, team
 
 teamRouter.get('/:team_id', teamMiddleware.validateTeamId, teamController.getTeam);
 
+teamRouter.patch('/:id',teamController.updateTeam);
+
```
aksh0001, line 16: I see a controller for updating a team, but do we have one for creating a team?

client/src/actions/users.js
```diff
@@ -7,10 +7,10 @@ import { ADD_USER_DATA } from "./types";
 /**
  * This action creator will be called when a user registers.
  *
- * @param userData object containing user email, given and family name and password
+ * @param userData object containing user email, team and organization name and password
  * @returns a thunk responsible for posting the data to the api and dispatching
  */
-export const addUserAction = (userData) => async (dispatch) => {
+export const addUserAction = (userData, apiResult) => async (dispatch) => {
```
aksh0001, line 13: Rather than taking an `apiResult` callback for signalling errors, I think it's better to dispatch separate actions (with different types) accordingly.
@pmyip3 did something similar: https://github.com/Researchify/Researchify/blob/scholar/client/src/actions/types.js#L22-L25

aksh0001, line 13: Rather than taking an `apiResult` callback for signalling errors, I think it's better to dispatch separate actions (with different types) accordingly.
@pmyip3 did something similar: https://github.com/Researchify/Researchify/blob/scholar/client/src/actions/types.js#L22-L25

aksh0001, line 13: Rather than taking an `apiResult` callback for signalling errors, I think it's better to dispatch separate actions (with different types) accordingly.
@pmyip3 did something similar: https://github.com/Researchify/Researchify/blob/scholar/client/src/actions/types.js#L22-L25

aksh0001, line 13: Rather than taking an `apiResult` callback for signalling errors, I think it's better to dispatch separate actions (with different types) accordingly.
@pmyip3 did something similar: https://github.com/Researchify/Researchify/blob/scholar/client/src/actions/types.js#L22-L25

client/src/actions/users.js
```diff
@@ -19,13 +19,15 @@ export const addUserAction = (userData) => async (dispatch) => {
...
       email: userData.email,
-      givenName: userData.givenName,
-      familyName: userData.familyName
+      teamName: userData.teamName,
+      orgName: userData.orgName
     };
 
     // Store user related data in our redux global store
     dispatch({ type: ADD_USER_DATA, payload: data });
+    apiResult(true);
```
aksh0001, line 28: same as above, prefer to dispatch an action that signals a failure.

client/src/actions/users.js
```diff
@@ -19,13 +19,15 @@ export const addUserAction = (userData) => async (dispatch) => {
...
   } catch (err) {
     console.error(`Error in posting user data to the api: ${err}`);
+    apiResult(false);
```
aksh0001, line 31: same as above, prefer to dispatch an action that signals a failure.

client/src/components/layout/Header.js
```diff
@@ -13,7 +13,7 @@ import { Fragment } from "react";
  * @returns Header component to be rendered in Layout.js
  */
 const Header = ({title, urls}) => {
```
aksh0001, line 15: Static Analyzer warnings, unused imports.

aksh0001 (commented):
Looks good overall, though I feel like we can address some comments before merging.

client/src/components/layout/Header.js
aksh0001: I wasn't able to target the affected lines in my comment above, but the unused imports are in this file.

api/src/models/user.model.js
kdav108: `teamName` - name of the research group (likely chosen by the members)
`orgName` - the Organisation name the group is part of, like Monash university, CSL, etc.
I have updated this in the document you linked.

api/src/routes/team.js
kdav108: @ssha0054 I think we don't right? I think when they register we create a new User, not really a team.

client/src/actions/users.js
kdav108: Don't you think that would get messy, we'll have so many different actions then. Unless we make a common action for api success and failure?

api/src/controllers/team.js
kdav108: Not sure about that controller. We may look to do it in the next PR.

kdav108:
Update: Added functionality for adding team in database upon registering. There's a lot of bad code here, repetitions, design issues, code smells, but for the demo hope it's alright 😂

PR #103, opened 2021-05-21
Aws integration

rsun0013:
# Description 
The aws lambda integration function to call the aws lambda and the importPublication function/ api endpoint to import all the publications for a given author id, publications parsed and returned in json that corresponds to the publication schema in db(but without team id).

# Type of change 
- [ ] Bug fix
- [1] New feature
- [ ] Refactoring
- [ ] Documentation 

# Solution description
same as main description

# Tests 
All unit tests pass?
yes
Builds successfully?
yes

Attach screenshot of the expected result (e.g. UI of the website/ postman api call result/ database changes)
publications in a json format
Screenshot (349)

Provide steps in detail for reviewers to reproduce the expected result of the code changes.
type localhost:5000/publications/importPublication/{authorid}

# Relevant document/ link (if any) 
Any related documents could help reviewers to review this PR 

# Risk
Identity the risk level (low, medium, high) and indicate which component(s) will be affected if bugs exist.
low

PR #101, opened 2021-05-20
Pmyip3/page to import publications using g scholar ms

pmyip3:
# Description 
When the import publication button is clicked and after the profile link is entered(accept any values for now), the hardcoded publications in json file will be displayed, which is pretended to be imported by scholar api (to be implemented).
The user should be able to click on the ones they want to insert and then when they're done, they click a button that sends that data to the backend to do a mass insert into the db. 

# Type of change 
- [ ] Bug fix
- [X] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
We need the feature of importing publications as per the client request. 

# Solution description

# Tests 
All unit tests pass? N/A

Builds successfully? Yes 

Attach screenshot of the expected result (e.g. UI of the website/ postman api call result/ database changes)
Profile Link would accpet any random characters for now
image
After loading, a list of publications would be populated from the json file, a checkbox attahed to each of those
image
After the import button is clicked, the selected publications will be sent to db and displayed on the page 
image

Error message will pop up when there is error from the scholar api call 
image
* to test this feature, uncomment line 117-120 and comment out line 113-116
image

# Relevant document/ link (if any) 
N/A

# Risk
Low for this PR alone, will be integared to the actual google scholar api.

PR #111, opened 2021-05-23
Pmyip3/integrating importing pubs

pmyip3:
# Description 
Import publications form google scholar using the actual api instead of the hardcoded publications in json file. 
Integrate [PR101 ](https://github.com/Researchify/Researchify/pull/101)and [PR103](https://github.com/Researchify/Researchify/pull/103)

# Type of change 
- [ ] Bug fix
- [X] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
Currently, we uses hardcoded publications in a json file to pretend it's imported by scholar api. We'll need to use the actual scholar api to import publications. 

# Solution description

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

After entering a valid profile link, it should return the following 2 publications (currently hardcoded from backend, due to lack of time before the demo)
Valid link example: https://scholar.google.com/citations?hl=en&user=3tyGlPsAAAAJ
image

The following error would pop up for invalid link or length of author_id !== 12
image
image

The 2 publications will be added to the current publications list if confirming import 
image

https://github.com/Researchify/Researchify/pull/101
https://github.com/Researchify/Researchify/pull/103

# Risk
Medium. The import publication feature won;t work if bugs exist


vgoh1:
Sorry for not catching these earlier in your PR but a few things:
image

- Published typo
- I think we can actually remove 'title:' and 'author:' because they're kind of self explanatory, and I don't like how the second publication is under the checkbox cause the title is too long. Is there a work around to get the part of the title that overflows to go under the checkbox instead of the whole title?

pmyip3:
> Sorry for not catching these earlier in your PR but a few things:
> image
> 
> * Published typo
> * I think we can actually remove 'title:' and 'author:' because they're kind of self explanatory, and I don't like how the second publication is under the checkbox cause the title is too long. Is there a work around to get the part of the title that overflows to go under the checkbox instead of the whole title?

@vgoh1 Fixed issues

vgoh1:
Thanks for fixing those - I'll hit approve but just a few more nitpicky things:
image

Space before the colon in "Year published : 2018" should be removed like "Link:"
And speaking of "Link:", could you make the url a clickable hyperlink?
I think we can probably differentiate between the field and the value, that is "Year published" can be underlined so that it's not the same font style as the year - but this can happen as a future improvement, I'll put it in the improvements ticket on Trello
And lastly, static analyzer says there's some unused imports - please get rid of those.
Feel free to merge after you've dealt with the issues above. Thanks!

vgoh1 (approved):

PR #125, opened 2021-07-08
Integrate Authentications

kdav108:
# Description 
After our discussion with client, we no longer store a `User`, but a `Team`. So this PR changes family and given Name attributes in team model to team and organisation name. 

This PR also makes the login functionality integrate with backend so authentication (login and registering) fully work now.

Note: Password security and JWT features are out of scope for this PR. Also out of scope is error handling with password and confirm password fields do not match in Register page (this is to avoid making this PR too big)

# Type of change 
- [X] Bug fix
- [ ] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
The login page was not currently integrated to backend
Client requirement change means that `givenName` and `familyName` are no longer appropriate

# Solution description
Connected login to backend
Updated team model to use `teamName` and `orgName` instead.

Login in with wrong credentials:
Screen Shot 2021-07-08 at 6 00 40 pm

Registering with no api backend running:
Screen Shot 2021-07-08 at 6 01 14 pm

When successfully Logged In, you should see team name (in this case, 'Mon') printed on dashboard:
Screen Shot 2021-07-08 at 6 01 36 pm

1) Run `nom run start` in `client` and `api`
2) In login page, use username: Mon@Uni and password: UniMon to login and see if it works
3) ^ It won't work, password is MonUni, try that
4) Also, try registering a new team from the register page.
5) To check if register page handles api errors, stop the `api` service and try registering.

# Risk
Medium risk


client/src/actions/team.js
```diff
@@ -1,15 +1,70 @@
...
+ * Adds a new team to redux store and database. If api request fails the team state's requestFailed attribute will be 'true'
+ * @param teamInfo contains teamName, orgName and emailnot
+ */
+export const addTeamInfo = (teamInfo, result) => async (dispatch) => {
+  try {
+    const data = await api.addTeam(teamInfo);
+    const teamData = {
+      ...teamInfo,
+      teamId: data.data._id,
+      requestFailed: false,
```
riordanalfredo, line 26: I don't really get this part... just wondering why is it hard-coded like this (?)

client/src/actions/users.js
```diff
@@ -7,10 +7,10 @@ import { ADD_USER_DATA } from "./types";
 /**
  * This action creator will be called when a user registers.
  *
- * @param userData object containing user email, given and family name and password
+ * @param userData object containing user email, team and organization name and password
  * @returns a thunk responsible for posting the data to the api and dispatching
  */
-export const addUserAction = (userData) => async (dispatch) => {
+export const addUserAction = (userData, apiResult) => async (dispatch) => {
```
riordanalfredo, line 13: uhm.. what to do with this apiResult?

riordanalfredo: I think this method is not called yet: 
image
(it got commented as shown image above)

client/src/actions/users.js
```diff
@@ -19,13 +19,15 @@ export const addUserAction = (userData) => async (dispatch) => {
...
       email: userData.email,
-      givenName: userData.givenName,
-      familyName: userData.familyName
+      teamName: userData.teamName,
+      orgName: userData.orgName
     };
 
     // Store user related data in our redux global store
     dispatch({ type: ADD_USER_DATA, payload: data });
+    apiResult(true);
```
riordanalfredo, line 28: I think it is better to return a boolean rather than passing a function like this. Wdyt?

riordanalfredo, line 28: same goes here

client/src/components/auth/RegistrationForm.js
```diff
@@ -28,7 +32,15 @@
         setInputs({...inputs, [name]: value})
     };
 
-    const [validated, setValidated] = useState(false);
+    const registrationResult = (isSuccessful, errorMessage) => {
```
riordanalfredo, line 35: I realised something about this: passing registrationResult function to the Redux action is not recommended: https://redux.js.org/style-guide/style-guide#reducers-must-not-have-side-effects

riordanalfredo (dismissed):
The functionality LGTM! Great work! I picked up about passing a function that may affect the state outside of the `redux action` scope. To fix it, we can add a new state, or return a value (i.e. boolean) after dispatching it. Well, this is just a suggestion to follow best practices.

Please follow my comments above :)

api/src/controllers/team.js
```diff
@@ -76,9 +76,44 @@ async function addTeam(req, res) {
...
+    .catch((err) => res.status(400).json('Error: ' + err));
+}
+
+/**
+ * Handles a POST request to add a team on the endpoint /team.
+ * @param {*} req request object -  json object containing at least two fields - teamName and orgName.
+ * @param {*} res response object - updated team object
+ * @returns 201: returns updated team details
+ */
+async function addTeam(req, res) {
```
vgoh1, line 106: I think your rebase may have went wrong somewhere since the static analyzer said there's two addTeam functions: lines 106-111 and 71-76

client/src/actions/team.js
vgoh1: To add, if the api request fails, requestFailed will be true, but in the catch I don't see it being set the true?

client/src/actions/team.js
```diff
@@ -1,15 +1,70 @@
...
+      error: "",
+    };
+    dispatch({
+      type: ADD_TEAM,
+      payload: teamData,
```
vgoh1, line 31: I think this dispatch isn't modifying the state, is that intentional?

client/src/actions/team.js
```diff
@@ -1,15 +1,70 @@
...
+  FETCH_TEAM_INFO,
+  GET_TEAM_MEMBERS_BY_TEAM_ID,
+  CREATE_TEAM_MEMBER,
+  UPDATE_TEAM_MEMBER,
+  DELETE_TEAM_MEMBER,
+  ADD_TEAM,
+} from "./types";
+
+/**
+ * Adds a new team to redux store and database. If api request fails the team state's requestFailed attribute will be 'true'
```
vgoh1, line 17: A suggestion cause scholar did it for the imported publications, you could have a ADD_TEAM_FAIL and ADD_TEAM_SUCCESS dispatch action to differentiate between whether the api request fails or not?

client/src/actions/team.js
```diff
@@ -1,15 +1,70 @@
...
+    const teamData = data.data.team;
+    console.log(teamData);
+    const team = {
+      teamId: teamData._id,
+      email: teamData.email,
+      teamName: teamData.teamName,
+      orgName: teamData.orgName,
+      twitterHandle: teamData.twitterHandle,
+      repoCreated: teamData.repoCreated,
+      requestFailed: false,
```
vgoh1, line 56: Should requestFailed and error be in the initial team state in teamReducer.js?

client/src/actions/team.js
kdav108: my bad, that variable should not be there, I forgot to remove it

client/src/actions/users.js
kdav108: The api result was added in the commit in thunder which I just cherry picked, and the function is not being used because we do not create a User model anymore when registering, we just create a Team model.

kdav108: The only reason I left the adding of `apiResult` in this function while cherry picking was in case we go back to using User model, don't want to loose effort

client/src/components/auth/RegistrationForm.js
kdav108: Interesting read. I couldn't find in the guide whether it says not to use callbacks inside Redux action. They recommend not doing it in Reducers though. Regardless, I searched around and I think I found a better way to do it. Thanks for this

api/src/controllers/team.js
kdav108: yep, thanks for picking this up

client/src/actions/team.js
kdav108: What makes you say that? It seems to update the store since we can see the team name on Dashboard.

client/src/actions/team.js
kdav108: These are removed now, not needed

client/src/actions/team.js
kdav108: Yep I did initially get inspired by Scholar and tried do it that way. However, the difference is that Scholar used a different React component for the error, while in this case the error is displayed by the same component. Happy to go into further discussion on how I wasn't able to use it in my case, but better to do it over a call :)

kdav108: I'm going to try a different method to having a callback tho, let me try that

kdav108:
I somehow changed something in `api/package-lock.json` not realising what. I have pushed it, so see if you guys can run the app fine after doing a npm install

client/src/actions/team.js
vgoh1: don't worry, looks like i was misreading

vgoh1 (approved):
LGTM, but I did realize when you log in you have no way of logging out so maybe that's something we can work on next + implementing the updating team info on the team profile page

kdav108:
>  I did realise when you log in you have no way of logging out

Good point, completely forgot about that. Maybe the top right user image can be clickable and a drop down with a log out option can be added

PR #127, opened 2021-07-13
Global team data in redux and global API error-handling

riordanalfredo:
# Description 
## Redux Error Handling by using error message from API response
The implementation of global error handling to print error messages from back-end to front-end. At the moment, it is placed in the `App` as shown below:
```javascript
const errorMessage = useSelector((state) => state.main.error); // uses global Redux state
<ErrorToaster message={errorMessage} /> // a component that will print the toaster
```

I left two pieces of functions:

image

if your team decides to use specific error-handling (such as `IMPORT_FAIL` in `ImportedPublications`), use the `errorerrorActionCreator` method and pass an appropriate Redux action type.
 
> Don't forget to put `ErrorToaster` component at the appropriate layout, and replace that `errorMessage` with the proper state.

## Apply TeamID
In the `App.js` component, I replace getting user action to `getTeamInfo`. 
image
> Don't forget to replace it with JWT token later.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation 

# Please do global search `FIXME` to help double-check everything


vgoh1 (approved):

client/src/actions/team.js
```diff
@@ -12,44 +13,47 @@ import {
...
+  try {
+    const data = await api.loginTeam(teamCredentials);
+    console.log(data);
+    const teamData = data.data.team;
+    const team = teamDataAllocator(teamData);
+    dispatch({
+      type: ADD_TEAM,
+      payload: team,
+    });
+    history.push('/dashboard');
```
kdav108, line 53: Would changing page from here be consistent with React guidelines? I assumed it's something the React component should do

kdav108 (commented):
Looks really good to me! I think having a standard way of doing error handling is nice. I'm just wondering though, should the error displaying be part of the react components since they handle all the UI? I know that would make quite a noticeable change to what has been done, but just throwing it out there for a discussion.

client/src/actions/team.js
riordanalfredo: Thanks for pointing it out! 💯  I totally forgot to revert it back to the original from my little experiment. I update it in my new commit below. 😆

riordanalfredo:
> Looks really good to me! I think having a standard way of doing error handling is nice. I'm just wondering, though, should the error displaying be part of the react components since they handle all the UI? I know that would make quite a noticeable change to what has been done, but just throwing it out there for a discussion.

Good question! From my experience at my previous workplace, we put try-catch at Redux. In fact, we wrapped our API with [try-catch function](https://stackoverflow.com/a/11774924) (I tried to replicate it, but it was way pretty challenging with async & await - need time). This is because we want to have clean React Components. After all, there were Container components, Layout components, Styled components, etc. Another benefit is the **proximity**, where they are located at a fewer number of files ( they are under one `actions` directory) in comparison to React components files (quite a lot). 

IMO, We can definitely have `then().catch()` at the Container components. However, it depends on how comfortable & discipline are your team in keeping the consistency of using `then().catch()` when calling the APIs. Also, some Layouts might need to be fixed to accommodate this design pattern ([Container](https://reactpatterns.com/#container-component) & [Layout components](https://reactpatterns.com/#layout-component)).

client/package.json
```diff
@@ -26,7 +26,6 @@
     "react-twitter-widgets": "^1.10.0",
     "redux": "^4.0.5",
     "redux-thunk": "^2.3.0",
-    "reselect": "^4.0.0",
```
kdav108: I think this was removed my accident?

kdav108:
Right, gotcha. You make a good point, and I also remembered learning that we should catch errors as close to the place where they arise from. I think the principle was Recover Early?? But anyways, this looks good to me, after we add `reselect` back in.

riordanalfredo:
> Right, gotcha. You make a good point, and I also remembered learning that we should catch errors as close to the place where they arise from. I think the principle was Recover Early?? But anyways, this looks good to me, after we add `reselect` back in.

Never heard of it before, 😆 lol; thanks! I'll put it in my notes for future reference!  Also, thanks for being thorough; I'll be careful next time ;)

kdav108:
I might have just coined that principle then 😂

kdav108 (approved):

PR #129, opened 2021-07-17
Pmyip3 authenication 

pmyip3:
# Description 
Storing authentication token & refresh token in the httpOnly cookie after a user has login successfully. 
Clear both cookies after logout. 
The passwords are encrypted before inserting into db. 
- the team created previously with an unencrypted password won't be able to log in
- pls remember the password when register new team when testing 

# Type of change 
- [ ] Bug fix
- [X] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
Currently, the website only authenticates the user credentials but not sending the tokens back to the user, which are needed for the authorization in the next ticket. 

# Solution description
See description 

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

*** As the encrypted password had been implemented, the teams created before with unencrypted passwords stored in db no longer can log in. Please register a new team before testing the below steps. After register successfully, the page will redirect to the dashboard automatically, pls log out and use the credential log in again to continues with the below steps, otherwise, the token won't be sent from the backend. ( Will improve this in a separate ticket)

image

After login successfully the 2 tokens can be found in the Cookies section
image

After clicking the logout button on the side bar, both tokens should be cleared. 
image
image

See above 

Any related documents could help reviewers to review this PR  N/A

# Risk
High. team controller from backend, login and register will be  affected if bugs exist.


api/src/controllers/team.js
```diff
@@ -70,25 +74,47 @@ async function getTeam(req, res) {
...
+    const foundTeam = await Team.findOne({ email: req.body.email })
+    if (!foundTeam) {
+      return res.status(400).send('Incorrect email/password'); // user not found 
+    } 
+    if (await bcrypt.compare(req.body.password, foundTeam.password)){
+      const teamObj = foundTeam.toObject(); // converts a mongoose object to a plain object 
+      // remove sensitive data 
+      delete teamObj.password 
+      const accessToken = jwt.sign(teamObj, process.env.JWT_SECRET_1 || "JWT_SECRET_1", {
+        expiresIn: '15m'
```
vgoh1, line 94: Do you have a specific reason for the chosen expiry times for access and refresh tokens? I think the access token can probably be valid for longer like 1 hour cause what if they have a lot of publications to add manually? Also see the second link below.

Was reading out of curiosity and apparently [Google](https://cloud.google.com/apigee/docs/api-platform/antipatterns/oauth-long-expiration) suggests that refresh tokens should expire a bit after the access token, but then [this page](https://blog.cloud-elements.com/oauth-2-0-access-refresh-token-guide) shows otherwise.

vgoh1 (approved):
Functionality wise, works great and is very exciting for what's next!

api/src/controllers/team.js
```diff
@@ -70,25 +74,47 @@ async function getTeam(req, res) {
...
+      });
+      const refreshToken = jwt.sign(teamObj, process.env.JWT_SECRET_2 || "JWT_SECRET_2", {
+        expiresIn: '1y'
+      });
+      res.cookie('accessToken', accessToken, { 
+        httpOnly: true,
+        maxAge: 30000, // 5 mins
```
vgoh1, line 101: Also since the expiry times are hardcoded 4 times, maybe it'll be easier to maintain by putting them in a file in the config folder? So we only have to change 2 values instead of 4 in that file.

pmyip3: Sure, will be easier to manage

api/src/controllers/team.js
pmyip3: After the access token expired, a request along with the refresh token will be sent to the backend to get a new access token. 
So a user will be forced to log in again only if the refresh token expires (or cookies expires as we are using cookies as well). 
We can decide and change the expires times later stage cause I am thinking to make them as short as possible to test it in the next ticket.

client/src/actions/team.js
```diff
@@ -19,12 +19,14 @@ import { errorActionGlobalCreator } from '../error/errorReduxFunctions';
  * @param teamInfo contains teamName, orgName and email
  */
 export const addTeamInfo = (teamInfo) => async (dispatch) => {
-  try {
-    const data = await api.addTeam(teamInfo);
+  try{
+    const teamId = await api.addTeam(teamInfo);
+    console.log(teamId)
```
cyr17, line 24: probably would need to remove this before merging

cyr17 (approved):
From my previous PR experience, we need to remove a bunch of console logs  before merging

api/src/config/tokenExpiry.js
```diff
@@ -0,0 +1,7 @@
+
+const aceessTokenExpiry = '15m';
```
vgoh1, line 2: Spelling mistake here

api/src/controllers/team.js
```diff
@@ -8,10 +8,16 @@ const Team = require('../models/team.model');
...
 
+const bcrypt = require('bcrypt');
+
+const jwt = require('jsonwebtoken');
+
 const options = {
   headers: { Authorization: 'Bearer ' + process.env.TWITTER_BEARER_TOKEN },
 };
 
+const { aceessTokenExpiry, refreshTokenExpiry, accessTokenCookieExpiry, refreshTokenCookieExpiry } = require('../config/tokenExpiry');
```
vgoh1, line 19: Typo here

PR #130, opened 2021-07-17
Profile Editing, Theme Selecting and Layout Route

jfan0010:
# Description 
Implemented functionality of editing team profile. Changes in dashboard page UI. Apply layout routes in App.js. Changes the directory of some files. Implemented functionality of dispatching theme selection to database.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
The team profile page is an empty form that does not update any information to the backend.
App.js has too much react-bootstrap component in it.
Users weren't able to select their choice of themes.

# Solution description
Added function in team controller and actions to update team data to the backend.
Almost all react-bootstrap component in App.js is removed, routes with preset layout are applied in App.js.
Added functionality to prompt and store user's theme selection.

# Tests 
All unit tests pass?
N/A

Builds successfully?
Yes

Profile Page

Once logged in, in team profile page, user can change their team name, organisation name and email in that page. 

The button of selecting theme and its form in dashboard page is not linked to the backend. 

# Risk
Medium. Will have conflict.


client/src/actions/types.js
```diff
@@ -20,6 +20,7 @@ export const CREATE_BULK_PUBLICATIONS = 'CREATE_BULK_PUBLICATIONS';
 export const FETCH_TEAM_INFO = 'FETCH_TEAM_INFO';
 export const LINK_TEAM_TWITTER = 'LINK_TEAM_TWITTER';
 export const UNLINK_TEAM_TWITTER = 'UNLINK_TEAM_TWITTER';
+export const UPDATE_TEAM = 'UPDATE_TEAN';
```
vgoh1, line 23: Spelling mistake here

client/src/components/github.js
```diff
@@ -0,0 +1,35 @@
+const axios = require("axios");
```
vgoh1, line 1: IMO this file should be in the backend rather than the frontend doing it, and to add, I believe this should actually be a part of scholly - which is in my branch

client/src/components/dashboard/Dashboard.js
```diff
@@ -1,16 +1,75 @@
...
+              </Card.Body>
+              <p>Editor</p>
+            </Link>
+          </Card>
+          <Card>
+            <Link className="researchify-dashboard-card-link">
+              <Card.Body>
+                <BsServer className="researchify-dashboard-card-icons" />
+              </Card.Body>
+              <p>API Acess Manager</p>
```
vgoh1, line 53: What's this page going to be for?

client/src/components/dashboard/TemplateSelector.js
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  const updateForm = (form) => {
+    const { name, value } = form.target;
+    if (name === 'theme') {
+      let primaryColor;
+      let secondaryColor;
+      switch (value) {
+        case 'theme1':
+          primaryColor = '#419aee';
```
vgoh1, line 36: If these themes are going to be stored in the db, then you shouldn't need to pass the colors into setInputs(), just the theme id (ie theme1/theme2/theme3)?

client/src/actions/types.js
jfan0010: Fixed. Took me 5 mins to find the `N`. XD

client/src/components/dashboard/Dashboard.js
jfan0010: So before it was to direct users to a page where they can manage their tokens and stuff, like twitter token or github. But since we're not getting their GitHub token anymore, I just can't decide what to replace it and keep the icon for now.

client/src/components/github.js
jfan0010: Oh right, this file is created by @kdav108  in the Logiciel branch right before the PI2 demo. Shall I remove this file?

client/src/components/dashboard/Dashboard.js
vgoh1: Yeah, it can probably be removed because I can't think what else it'd be for either

client/src/components/github.js
vgoh1: I think leave it for now in case we have some discussion over it in my PR / meetings, and then we can delete it before you merge

client/src/components/dashboard/TemplateSelector.js
jfan0010: These colour combinations are not finalised yet and in our `theme.model` database, `themeId` will represent the combination of layout and colours. We don't have a individual model to store colour combinations. If we're to give each colour combination an id, I reckon we might need to manually put them into the database? The reason I put it here is mainly because we can change them from the front end and create a new theme in database through API calls.

client/src/components/dashboard/Dashboard.js
jfan0010: Removed, it looks kinda weird when there're only 2 icons under tho XD

api/src/controllers/theme.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {*} req 
+ * @param {*} res 
+ * @returns 
+ */
+async function findOrCreateTheme(req, res) {
+  const themeData = req.body;
+  try {
+    const foundTheme = await Theme.findOne(themeData);
+    if (!foundTheme) {
+      const createdTheme = await Theme.create(themeData);
```
pmyip3, line 18: Just want to clarify, why the user is able to select a theme that is not currently existing in db? 
Do we allow them to choose their custom color combination?

jfan0010: No, just because we have 3 layouts and 3 colour combinations to choose from, and we don't want to put them into the database manually before users selecting their themes. Also if we decided to change or add more options, there will be more combination to put in the database.

client/src/components/dashboard/TemplateSelector.js
vgoh1: Yeah, otherwise you would have to insert them manually into the database anyway. This is fine for now then!

api/src/controllers/theme.js
pmyip3: Got it !

pmyip3 (approved):

vgoh1 (approved):
LGTM, just remember to remove the file I mentioned before merging. :)

cyr17 (approved):
all good

PR #132, opened 2021-07-21
Refactoring backend for better error handling

vgoh1:
# Description 
Try and catch blocks have been refactored to use promises instead, to facilitate the use of express' default error handler. Also improved the error messages to follow a certain format so that it's easier to handle on the frontend. Removed endpoints that weren't being used (Read a single publication)

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
See https://trello.com/c/DwKSFNmj/68-refactor-backend

# Solution description
There's a function fillErrorObject that is called whenever an error occurs, to put the error in a nicer format:
`const errObject = {
      errorCode: code,
      errorMessage: msg,
      errors: [...err],
      errorTimestamp: localISOTime
    };`
Error handling middleware has also been added to server.js so it goes to that if next(err) is used.

# Tests 

Hit any of the endpoints in postman and the error should look like so:
image

The easiest one to test on the frontend is login with incorrect password, you'll see the toast and error state change:
image

# Risk
High, all of backend will be affected


api/src/server.js
```diff
@@ -33,5 +33,17 @@ app.use('/publications', publicationsRouter);
...
 app.use('/theme', themeRouter);
 
+// error handler middleware
+app.use(function (err, req, res) {
+  const errorObject = err;
+  if (errorObject) {
+    res.status(err.errorCode).json(err);
+  } else {
+    // if error object is not passed, server error
+    res.status(500).send('Something went wrong!');
```
kdav108, line 43: Do we need to change this to something else other than 'Something went wrong'?

api/src/middleware/mongoose.js
```diff
@@ -0,0 +1,28 @@
...
+  if (!mongoose.Types.ObjectId.isValid(_id)) {
+    console.log('in here');
+    console.log(_id);
+    next(
+      fillErrorObject(400, 'Validation error has occurred', [
+        'Given ' + type + ' id is not in a valid hexadecimal format',
+      ])
+    );
+  } else {
+    next();
```
kdav108, line 14: Does this not go to the error function?

api/src/middleware/mongoose.js
```diff
@@ -0,0 +1,28 @@
+const mongoose = require('mongoose');
+const { fillErrorObject } = require('../middleware/error');
+
+async function validateObjectId(_id, type, res, next) {
+  if (!mongoose.Types.ObjectId.isValid(_id)) {
+    console.log('in here');
+    console.log(_id);
+    next(
+      fillErrorObject(400, 'Validation error has occurred', [
```
kdav108, line 9: It's a nit pick, but I think convention is to just use 'Validation error', i.e. no need for a proper sentence. But it's a really small change so don't think it would be worth going through all these and changing them?

kdav108 (commented):
Nice small PR! Looks good to me, just a few minor things I noted

api/src/controllers/publications.js
```diff
@@ -21,19 +23,23 @@ const { puppeteerConfig, categoryType } = require('../config/puppeteer');
...
-
-  try {
-    await Publication.findByIdAndRemove(_id);
-    return res
-      .status(200)
-      .json({ message: 'Publication deleted successfully.' });
-  } catch (err) {
-    res.status(400).json(`Error: ${err.message}`);
-  }
+  await Publication.findByIdAndRemove(_id)
```
pmyip3, line 28: If using promises instead, I don't think we will need to await a promise and the function no longer need to be async. Same for the other functions as well.

pmyip3 (changes requested):

api/src/server.js
vgoh1: Yes we do, thanks for the reminder!

api/src/middleware/mongoose.js
vgoh1: Yes, because there's no argument passed to next(), so it won't go to the error function, it'll go on to run the specified function in the controller file. If it passes the objectId validation, then there's no need to go to the error function.

api/src/middleware/mongoose.js
vgoh1: You're right, validation error does sound better. No worries, I'll go change it, that's what replace all is for. :)

api/src/controllers/publications.js
vgoh1: I tested it and looks like you're right! I'll change them accordingly then.

api/src/server.js
```diff
@@ -33,5 +33,18 @@ app.use('/publications', publicationsRouter);
 app.use('/team', teamRouter);
 app.use('/theme', themeRouter);
 
+// error handler middleware
+app.use(function (err, req, res, next) {
```
vgoh1, line 37: It doesn't work without next in here so ignore the static analyzer....

pmyip3 (approved):

jfan0010 (approved):
LGTM. Btw if you need to include  `next` as the parameter of function in `server.js`, ig `// eslint-disable-next-line no-unused-vars` could help you get rid of the static analyser warning if you add it right before `line 37` in `api.src.server.js`

kdav108:
LGTM!

PR #133, opened 2021-07-23
Pmyip3 authorization

pmyip3:
# Description 
Authorization functionality built on both backend and front end:
Backend: auth middleware is implemented to ensure the user has valid tokens before access to protected endpoints 
Frontend: private route can only be accessed in frontend if after a user login successfully 

# Type of change 
- [ ] Bug fix
- [X] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
Currently, any user can access any endpoints and route without authorization, which is a problem.  

# Solution description
Based on the previous authentication PR, after a user login successfully, there are 2 HTTP-only cookies (access and refresh token) that will be sent from the backend. When users try to make protected endpoints requests, these two tokens will be sent along to the backend to ensure the request is authorized. 

A new cookie(not HTTP-only so that we can access the value in the front end) is created in this PR, named isLogin so we can check if the user is currently login even if the page is refreshed (as the state will be gone). 

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

1. Before login, attempts to access private route e.g. locallost:3000/dashboard will be redirected to the  locallost:3000
2. After an access token expires, the new access token should be automatically generated given the refresh token hasn't expired. Feel free to test this by shortening the expiration time in api/src/config/tokenExpiry.js
3. If both access and refresh tokens expire and access/ refresh token does not exist when the user makes a request to a protected endpoint, an error would be thrown. 

Note: both cookies expiry values should be the same or longer than the refresh token expiry, whereas the access token expiry should be less than the above three. 

# Risk
Medium. If bugs exist only affect the authorization functionality itself but not other components


api/src/config/tokenExpiry.js
```diff
@@ -1,7 +1,6 @@
+const aceessTokenExpiry = '15m';
```
vgoh1, line 1: This is a nitpick, but I almost wonder if we can pull out the value in minutes, and then for the token expiry convert it to string and add a 'm' at the end. So that this value can be used for the cookie expiry as well by multiplying it by 60000 milliseconds per minute. eg:
`const accessTokenExpiryValue = 15;`
`const accessTokenExpiry = accessTokenExpiryValue.toString() + 'm';`
`const accessTokenCookieExpiry = accessTokenExpiryValue * 60000;`
That way we only need to change accessTokenExpiryValue, and not accessTokenExpiry and accessTokenCookieExpiry.

vgoh1: Also spelling mistake in access

api/src/controllers/auth.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+const Team = require('../models/team.model');
+
+const bcrypt = require('bcrypt');
+
+const jwt = require('jsonwebtoken');
+
+const { fillErrorObject } = require('../middleware/error');
+
+const { aceessTokenExpiry, refreshTokenExpiry, accessTokenCookieExpiry, refreshTokenCookieExpiry } = require('../config/tokenExpiry');
```
vgoh1, line 13: FYI spelling mistake in access here too

api/src/controllers/auth.js
```diff
@@ -0,0 +1,87 @@
...
+    if (await bcrypt.compare(req.body.password, foundTeam.password)){
+      const jwtPayload = {
+        _id: foundTeam._id,
+        email: foundTeam.email,
+        teamName: foundTeam.teamName,
+        orgName: foundTeam.orgName
+      }
+
+      const accessToken = jwt.sign({ team: jwtPayload }, process.env.JWT_SECRET_ACCESS_TOKEN || "JWT_SECRET_ACCESS_TOKEN", {
+        expiresIn: aceessTokenExpiry
```
vgoh1, line 41: FYI spelling mistake

api/src/controllers/auth.js
```diff
@@ -0,0 +1,87 @@
...
+
+    if (await bcrypt.compare(req.body.password, foundTeam.password)){
+      const jwtPayload = {
+        _id: foundTeam._id,
+        email: foundTeam.email,
+        teamName: foundTeam.teamName,
+        orgName: foundTeam.orgName
+      }
+
+      const accessToken = jwt.sign({ team: jwtPayload }, process.env.JWT_SECRET_ACCESS_TOKEN || "JWT_SECRET_ACCESS_TOKEN", {
```
vgoh1, line 40: Out of curiosity, do you actually have a JWT_SECRET_ACCESS_TOKEN that you're using or is it defaulting to the string?

pmyip3: No, haven't created it yet, I am just using the default string here.

api/src/config/tokenExpiry.js
pmyip3: I think I made a serious mistake here cause the accessTokenExpiry should not be equal to its accessTokenCookieExpiry. Even though the accessToken is expired, we still need this to pass it to the backend to generate a new one, so we don't want it to be removed from the cookies section. I'll get it fixed.

jfan0010 (commented):
When the token is expired, the user can still change the team info in profile page and navigate between pages until log out manually. I think redirecting or log out the user to the homepage/landing page once the token is expired might be better?
Also, found a problem that probably is not related to this PR. Once the token is expired, whenever I tried to log in / log out (succeed or not), the toast error message will show the message of token expired.

pmyip3:
> When the token is expired, the user can still change the team info in profile page and navigate between pages until log out manually. I think redirecting or log out the user to the homepage/landing page once the token is expired might be better?
> Also, found a problem that probably is not related to this PR. Once the token is expired, whenever I tried to log in / log out (succeed or not), the toast error message will show the message of token expired. (Could probably raise an issue later in semester ig)

The fix bug should be solved now. However, found another bug in the profile page: an error message won't occur if there is an error from the API request> I have done a quick fix by dispatch a gobalError Action. However, a successfully updated message won't be shown now, might need to raise another ticket to address this. 

For the second point, I have made a change to dispatch a CLEAR_ERROR action after signout so that the user won't receive any error message from the previous message ( as it might still in the redux store if the page doesn't get refreshed).

vgoh1 (approved):

cyr17 (approved):

api/src/config/tokenExpiry.js
```diff
@@ -1,7 +1,12 @@
+// Note: both cookies expiry values should be the same or longer than the refresh token expiry
+// whereas the access token expiry should be less than the above three.
+
+const refreshTokenExpiryValue = 10080; // 7 days in minute 
+const cookiesExpiryValue = refreshTokenExpiryValue
 
 const accessTokenExpiry = '15m';
```
vgoh1, line 7: I think a comment to explain what changing this does would be nice

PR #135, opened 2021-07-28
Introduce the `Website` schema to store client website metadata

kdav108:
# Description 
This PR adds a `Website` schema which holds information about the client website including url name, the pages the client has added and the title of the website.

3 API endpoints - get, add page, deleted page - are added to interact with the schema for now

This PR changes both the front and backend - a replica of the schema is added into Redux store.

Note: A few things still need to be worked on but are left for future PRs, these include:
- When a user closes the select theme pop up, it currently goes on to show the add page pop up. This shouldn't be the case, selecting the theme is a requirement.
- The theme selection is not currently in any of the redux store, hence the do point before cannot yet be implemented inn a nice and easy way.
- Based on the pages array of the website schema, we need to tell schools to only deploy those pages (e.g. although we may have data about client's publications, if they have deleted the PUBLICATIONS page, we do not currently remove their publications data - and I don't think we should - so we need to tell scholly to avoid pushing publications page to client website
- An edit button should appear on the home page with each client PAGE list displayed. This should probably replace the pages on the sidebar

# Type of change 
- [ ] Bug fix
- [X] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
We had no way of holding data about client's website (e.g. the pages they wanted published)

# Solution description
See description above (introduction of website schema)

When logged in as SomeTeam:
Screen Shot 2021-07-28 at 8 19 07 pm

Adding a page:
Screen Shot 2021-07-28 at 8 19 16 pm

Deleting the TEAM Page:
Screen Shot 2021-07-28 at 8 19 32 pm

1) Run both client and server
2) Login as a test user 
3) Play around by adding and deleting pages from the DAShboard
4) Confirm the change appears in UI and in DB

A design of the website schema can be found [here](https://www.notion.so/Researchify-DB-Schema-78d164ce0c6147f9831f79cd9b89c519)

# Risk


api/src/controllers/website.js
```diff
@@ -0,0 +1,134 @@
+/**
+ * This module contains handlers for the "clientWebsite" route.
+ * @module website
+ */
+const axios = require('axios');
```
aksh0001, line 5: Please fix Static Analyzer checks

api/src/controllers/website.js
```diff
@@ -0,0 +1,134 @@
...
+
+const Website = require('../models/website.model');
+
+const { fillErrorObject } = require('../middleware/error');
+
+var ObjectId = require('mongoose').Types.ObjectId;
```
aksh0001, line 11: use `const` instead of `var`: https://ui.dev/var-let-const/

api/src/controllers/website.js
```diff
@@ -0,0 +1,134 @@
...
+ */
+async function deleteWebPage(req, res, next) {
+  const { team_id } = req.params;
+  Website.findOne({ teamId: team_id })
+    .then((website) => {
+      if (website == null) {
+        res.status(400).send('This team does not have a website created');
+      } else {
+
+        index = website.pages.indexOf(req.body.pageName);
```
aksh0001, line 115: Please fix SA checks

api/src/controllers/website.js
aksh0001: Also, this variable may not need to be global. Prefer reducing the scope.

api/src/controllers/website.js
```diff
@@ -0,0 +1,134 @@
...
+  Website.findOne({ teamId: info.teamId })
+    .then((website) => {
+      if (website == null) {
+        // Create Website data
+        Website.create(info).catch((err) => {
+          throw err;
+        });
+      }
+    })
+    .catch((err) => {
```
aksh0001, line 55: Is there a reason we're using `.then().catch()` i.e. promise style here, but async-await style elsewhere? Maybe we should stick to one convention

aksh0001 (approved):
Minor changes needed, but otherwise lgtm!

api/src/controllers/website.js
kdav108: Fair point, I've converted majority into promise-style as this seems to be more common in other files as well.

api/src/controllers/website.js
kdav108: Turns I don't need ObjectId all together, so removing it... 😂

kdav108:
Update: 
- I have added an Edit button to Dashboard as well (I was leaving it for later but turned out to be quite easy).
- Hence, I have removed the Editor page from the sidebar (It still exists in our code, in case we need it in the future).
- Also, I have made the sidebar dynamic such that Publications and Team Member page is only shown if the client has added those pages from the Dashboard. Let me know if you think we should always have them in the sidebar, and I'm happy to revert it back.

rpra0006 (approved):

client/src/actions/auth.js
```diff
@@ -2,68 +2,94 @@
...
+  } catch (error) {
     dispatch({
-      type: AUTH_SIGN_IN_FAIL
-    })
+      type: AUTH_SIGN_IN_FAIL,
+    });
     dispatch(errorActionGlobalCreator(error));
   }
+
+  await api
```
pmyip3, line 44: I don't it is a proper way to put the getWebsiteInfo api request here as it will only be called when the user first log in, and once the user refreshes the page after login, all the info stored in the redux state would be gone and it is no way to retrieve this info again unless the user log in again.

client/src/actions/website.js
```diff
@@ -1,23 +1,65 @@
...
+
+/**
+ * Deletes info about a client's web-page from redux store and database.
+ * @param teamInfo contains teamName, orgName and email
+ */
+export const deletePage = (teamId, pageName) => async (dispatch) => {
+  const apiBody = {
+    pageName: pageName,
+  };
+  await api.deleteWebPage(teamId, apiBody).then(
```
pmyip3, line 59: Is this line mixes await/async and promises?

pmyip3 (changes requested):

client/src/actions/website.js
kdav108: hahah yeah don't know how I missed it

client/src/actions/auth.js
kdav108: True... where would you think this should be placed to avoid data being deleted when the page is reloaded. I'm thinking we can have it in the dashboard, but wanted to hear your thoughts @pmyip3

pmyip3: Yes, we could put the API request inside an useEffect hook in the Dashboard component, so every time the dashboard page gets refreshed, the request will be sent to the backend again. But even if we are not on the Dashboard page and we try to refresh the page, we still want to send that API request as the relevant pages should be on the sidebar. 

So I think we can just also return the website Info in the authoorizeJWT function as this function will get called every time the private route gets refreshed. 
image


In this way, the getTeamJWT API request would return both team_info and website_info, then we can just dispatch two actions with the corresponding results. 
image

pmyip3: The corresponding backend endpoints for getTeamJWT() need to query both the teams and website collection as well in order to return both information. 

Just my thought, might require a lot of changes though. Feel free to comment on it.

vgoh1:
Some general comments on the functionality:
- You're going to need some validation on whether the pages have already been added to the document or else this will happen in the DB: 
image
- If the page has already been added, they should show up on the dashboard already
- I do like how you actually remove the page from the sidebar if it's not on the dashboard, however this message: `Deleting the page would permanently remove all data associated with this page and the page will not be shown on your website` is a bit misleading I think in that it wouldn't 'permanently remove all data associated with this page' cause the data is still in the DB
- I feel like selecting the color combination and layout should only happen once for the entire site as we had discussed with the client in class last week, so maybe you can pull that up as a seperate button or something on the dashboard
- If all available pages have been added, then the user shouldn't be able to get to the 'choose the page to add to your website' screen

Please add the points you mentioned in the inbox on the trello board so we can keep track of them, but I do have a question about this:
- `An edit button should appear on the home page with each client PAGE list displayed. This should probably replace the pages on the sidebar` <- What do you mean by this? Cause I personally like the fact that they show up on the sidebar, since I don't want the edit button to be the only way I can get to the page.

client/src/reducers/websiteReducer.js
```diff
@@ -1,23 +1,50 @@
...
+    ADD_WEBPAGE,
+    DELETE_WEBPAGE,
+    FETCH_WEBSITE_INFO,
+  } from '../actions/types';
+  
+  const INITIAL_WEBSITE_STATE = {
+    url: '',
+    title: '',
+    pages: [],
+    availablePages: ['PUBLICATIONS', 'TEAM'],
```
vgoh1, line 15: You can pull the availablePages up into /config so it'll be easier to maintain when we need to add more pages

kdav108: Great idea! I didn't know config existed lol

kdav108:
>An edit button should appear on the home page with each client PAGE list displayed. This should probably replace the pages on the sidebar <- What do you mean by this? Cause I personally like the fact that they show up on the sidebar, since I don't want the edit button to be the only way I can get to the page.

I agree actually

>If all available pages have been added, then the user shouldn't be able to get to the 'choose the page to add to your website' screen

I have added a toaster for the time being, although I think it's not a great way to tell the user there's no more pages to add, I think we can refactor it later

vgoh1 (approved):

client/src/actions/auth.js
```diff
@@ -37,33 +53,53 @@ export const signIn = (authData) => async(dispatch) => {
...
+    });
+  } catch (err) {
     dispatch(errorActionGlobalCreator(err));
   }
 };
 
-export const authorizeJWT = () => async(dispatch) => {
-  try{
-    const { data } = await api.getTeamJWT()
+export const authorizeJWT = () => async (dispatch) => {
```
pmyip3, line 71: Just one thing can improve performance, as both sidebar info and team info are needed to get from the backend for every refreshing, maybe we can just combine them to send only one API call instead of two separate calls? 
We can improve this in a separate ticket though.

pmyip3 (approved):
Other than the performance issue which we might need to consider in a later stage, looks good to me for the remaining. Thanks for the work.

jfan0010 (approved):
LGTM! I can remove those editor-related files later if we're not gonna reuse them at all.

cyr17 (approved):

api/src/controllers/website.js
```diff
@@ -0,0 +1,142 @@
...
+        return res
+          .status(404)
+          .send(
+            'No Website info found for this Team (This is likely because they have not added any website information yet)'
+          );
+      }
+      return res.status(200).send(websiteInfo);
+    })
+    .catch((error) => {
+      return res.status(500).json(`Server Error: ${error.message}`);
```
aksh0001, line 33: Can this be turned to async-await to make the code cleaner and avoid the nesting?

pmyip3: We can do the refracting in a separate ticket. Will merge this to main for now.

aksh0001: Was the commnent by @vgoh1 resolved?
https://github.com/Researchify/Researchify/pull/135#issuecomment-890447160
@kdav108

PR #136, opened 2021-07-28
Backend deploy fix

vgoh1:
# Description 
Before the deploy to scholly endpoint was using a hardcoded teamid due to the Github redirect back to Researchify losing all login information. Now with access tokens properly implemented, this is no longer the case and the teamid can be retrieved from the state after the redirect.

# Type of change 
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
See description

# Solution description

# Tests 
Make a new account, add some publications, and then go to the deploy page to deploy.
You can put a `console.log(body)` in /api/controllers/team.js after line 225 to check whether the publications you just added are being passed through properly.

N/A

# Risk
Low, just the deploy flow will be affected

### DIFF
--- a/api/src/controllers/team.js
+++ b/api/src/controllers/team.js
@@ -34,7 +34,6 @@ const options = {
 async function storeHandle(req, res, next) {
   const { twitterHandle: handle } = req.body;
   let foundTeam = req.foundTeam;
-  console.log('in here');
 
   if (handle.length == 0) {
     // remove the handle from the doc
--- a/api/src/controllers/team.js
+++ b/api/src/controllers/team.js
@@ -223,16 +223,14 @@ async function deployToGHPages(req, res, next) {
     teamPublications: publications,
   };
 
-  await axios({
-    url: schollyHost + '/deploy/' + teamId,
-    method: 'post',
-    data: body,
-    headers: {
-      'Content-Type': 'application/json',
-    },
-  })
+  await axios
+    .post(`${schollyHost}/deploy/${teamId}`, body)
     .then(() => res.status(200).json('Successfully deployed'))
-    .catch(() => next(fillErrorObject(500, 'Server error', ["Error occurred with scholly"])));
+    .catch(() =>
+      next(
+        fillErrorObject(500, 'Server error', ['Error occurred with scholly'])
+      )
+    );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


aksh0001 (approved):
Approving for general readability. I'll let someone else handle the core business logic as I'm still getting up to speed with the new changes

pmyip3 (approved):

client/src/components/deploy/DeployPage.js
```diff
@@ -10,26 +10,28 @@ const DeployPage = () => {
   // const teamId = useSelector((state) => state.team.teamId);
```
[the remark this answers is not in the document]
vgoh1: Done, totally forgot about it, thanks!

cyr17:
So I'm getting a popup saying all publications have been imported while trying to import publications.

I tried importing using a random tutors profile link and I still get the same pop-up


image


this is what I get in the terminal I assume it's because of the long duration
image

api/src/controllers/team.js
```diff
@@ -220,19 +223,16 @@ async function deployToGHPages(req, res) {
...
-      headers: {
-        'Content-Type': 'application/json',
-      },
-    });
-    return res.status(200).json(schollyResponse.data);
-  } catch (err) {
-    console.log(err);
-  }
+  await axios({
+    url: schollyHost + '/deploy/' + teamId,
```
aksh0001, line 227: Prefer template literals here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

api/src/controllers/team.js
```diff
@@ -220,19 +223,16 @@ async function deployToGHPages(req, res) {
...
+    method: 'post',
+    data: body,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  })
```
aksh0001, line 233: This may not be as readable as a simple `axios.post`. The json content type will be automatically applied if the body is a JS object: https://masteringjs.io/tutorials/axios/post-json

aksh0001 (changes requested):
Maybe we could address those style-related changes?

aksh0001 (approved):
LGTM, Approval

jfan0010 (approved):
Works fine on mine! The GH page is updated as expected when deploy again. And we can add a loading icon when deploying and a toast message to tell users when the page is deployed in later stage :)

PR #138, opened 2021-07-31
Import publications fix

vgoh1:
# Description 
Import publications would time out in the backend for some reason. After investigating, I found out that Google has changed the HTML of the individual publication pages, breaking the scraper. Additionally, sometimes it would take an abnormally long time to load even the profile (30s~) causing Puppeteer to time out.

# Type of change 
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
See description.

# Solution description
The HTML tags it scrapes from have been updated, but also instead of using Puppeteer which only has Chrome support, it's been changed to Playwright instead (made from the same team so the logic is mostly the same) so we have access to Firefox which will be more lightweight (thanks Akshay for the suggestion). 

The scraping logic has been moved to its own function (scrapeGoogleScholar), while the initial setup like getting the individual publication urls on google scholar is in getGoogleScholarPublications.

# Tests 
As usual, feel free to play around with this and break it so we can iron out the bugs
**If testing in postman:** 
Comment out the auth middleware for the endpoint if you're too lazy to put a cookie in postman like me (thanks Mable)
Do a GET to this endpoint: `http://localhost:5000/publications/import/{gScholarUserId}/{startFrom}/validate/{teamId}`
eg: `http://localhost:5000/publications/import/eRbvWqYAAAAJ/10/validate/609f5ad827b1d48257c321d3`
startFrom is used to facilitate pagination, so eg if we got the first 10 publications already, we would start from 10 to get the next publications
teamId is used to validate the imported publications whether they're in the database already

**If testing via frontend:**
Go to publications page -> click import
Keep in mind there are two bugs:
1. If you get the first ten publications, and import them, and then click on import, it will say you've imported all the publications. This is because the current logic assumes that the user will only import once, hence if the latest publications are in the db already then everything else before that should be imported. I want to improve this logic in a later ticket so if anyone has any suggestions please let me know.
2. Pagination is prone to breaking if click to go to another page that isn't next in order to your current page (eg going from page 5 to page 1) where no publications will show up. If you use the arrow buttons to navigate the pages this doesn't happen. If anyone wants to fix this for me, let me know. :) Or else I'll fix this later.

# Risk
Low, should just affect the import functionality

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -24,8 +24,6 @@
     "jsonwebtoken": "^8.5.1",
     "mongoose": "^5.12.2",
     "playwright": "^1.13.1",
-    "puppeteer": "^10.0.0",
-    "puppeteer-cluster": "^0.22.0",
     "validator": "^13.5.2",
     "winston": "^3.3.3"
   },
--- a/api/package.json
+++ b/api/package.json
@@ -23,6 +23,7 @@
     "express-validator": "^6.10.0",
     "jsonwebtoken": "^8.5.1",
     "mongoose": "^5.12.2",
+    "playwright": "^1.13.1",
     "puppeteer": "^10.0.0",
     "puppeteer-cluster": "^0.22.0",
     "validator": "^13.5.2",
--- a/api/src/config/playwright.js
+++ b/api/src/config/playwright.js
@@ -4,7 +4,6 @@
 
 const playwrightConfig = {
   noOfDummyLinks: 4, // first 4 links aren't publications can be ignored
-  noOfThreads: 10, // number of concurrent threads
   pageSize: 10, // number of publications to return at once
   gScholarHome: 'https://scholar.google.com.sg',
   baseUrl: 'https://scholar.google.com.sg/citations?hl=en&user=',
--- a/api/src/config/playwright.js
+++ b/api/src/config/playwright.js
@@ -2,11 +2,11 @@
  * This module stores the configuration for using Puppeteer to scrape Google Scholar.
  */
 
-const puppeteerConfig = {
+const playwrightConfig = {
   noOfDummyLinks: 4, // first 4 links aren't publications can be ignored
   noOfThreads: 10, // number of concurrent threads
   pageSize: 10, // number of publications to return at once
-  base: 'https://scholar.google.com.sg',
+  gScholarHome: 'https://scholar.google.com.sg',
   baseUrl: 'https://scholar.google.com.sg/citations?hl=en&user=',
   startSuffix: '&cstart=', // concatenate number to start at after
   pageSizeSuffix: '&pagesize=', // concatenate page size after
@@ -22,4 +22,4 @@ const categoryType = {
 
 const categoryTypeEnum = ['CONFERENCE', 'JOURNAL', 'OTHER', 'BOOK'];
 
-module.exports = { puppeteerConfig, categoryType, categoryTypeEnum };
+module.exports = { playwrightConfig, categoryType, categoryTypeEnum };
--- a/api/src/config/playwright.js
+++ b/api/src/config/playwright.js
@@ -5,8 +5,8 @@
 const playwrightConfig = {
   noOfDummyLinks: 4, // first 4 links aren't publications can be ignored
   pageSize: 10, // number of publications to return at once
-  gScholarHome: 'https://scholar.google.com.sg',
-  baseUrl: 'https://scholar.google.com.sg/citations?hl=en&user=',
+  gScholarHome: 'https://scholar.google.com',
+  baseUrl: 'https://scholar.google.com/citations?hl=en&user=',
   startSuffix: '&cstart=', // concatenate number to start at after
   pageSizeSuffix: '&pagesize=', // concatenate page size after
   sortBySuffix: '&view_op=list_works&sortby=pubdate', // put most recent pubs first
--- a/api/src/controllers/publications.js
+++ b/api/src/controllers/publications.js
@@ -8,6 +8,8 @@ const Publication = require('../models/publication.model');
 
 const Team = require('../models/team.model');
 
+const logger = require('winston');
+
 const { firefox } = require('playwright');
 
 const { playwrightConfig, categoryType } = require('../config/playwright');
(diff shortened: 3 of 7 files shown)


api/src/models/publication.model.js
```diff
@@ -52,8 +52,6 @@ const publicationSchema = new mongoose.Schema(
       },
       categoryTitle: {
         type: String,
-        required: true,
```
vgoh1, line 55: This was removed because the user might not have given a category title when adding their publications to google scholar

api/package.json
```diff
@@ -23,8 +23,7 @@
     "express-validator": "^6.10.0",
     "jsonwebtoken": "^8.5.1",
     "mongoose": "^5.12.2",
-    "puppeteer": "^10.0.0",
-    "puppeteer-cluster": "^0.22.0",
```
aksh0001, line 27: Curious which PR you intend to submit first. This or https://github.com/Researchify/Researchify/pull/136 ?
The latter relies on `puppeteer` deps.

api/src/config/playwright.js
```diff
@@ -2,10 +2,11 @@
  * This module stores the configuration for using Puppeteer to scrape Google Scholar.
  */
 
-const puppeteerConfig = {
+const playwrightConfig = {
   noOfDummyLinks: 4, // first 4 links aren't publications can be ignored
   noOfThreads: 10, // number of concurrent threads
```
aksh0001, line 7: Can the number of threads be set more dynamically than a literal value? Do you know if there's a way to find the optimal thread count given hardware limitations?
In a C/C++ multirhreaded app, you could call `std::thread::hardware_concurrency()` to find the max number of threads that can run in parallel. Anything more will degrade performance due to constant thread switching

aksh0001:
Also, please update this branch with `main`'s latest changes :)

aksh0001 (changes requested):
I'm curious to see if there's a better way to determine optimal thread count for playwright

api/src/config/playwright.js
vgoh1: Actually, I don't think I'm using noOfThreads anymore and forgot to remove it. Right now it's using promises to scrape in 'parallel' in line 175 of controllers/publications.js

api/package.json
vgoh1: #136 will go in first so there's less conflicts to deal with

api/src/config/playwright.js
vgoh1: Playwright as far as I saw doesn't have anything in built to handle parallelism so I used the solution mentioned here: https://github.com/microsoft/playwright/issues/4729

api/src/controllers/publications.js
```diff
@@ -270,6 +192,72 @@ async function getGoogleScholarPublications(req, res) {
   res.status(200).json(response);
 }
 
+async function scrapeGoogleScholar(url) {
+  console.log(playwrightConfig.gScholarHome + url);
```
aksh0001, line 196: Please use the `winston` logger for logging so we can inspect logfiles for debugging in AWS.

api/src/controllers/publications.js
```diff
@@ -144,118 +144,40 @@ async function getGoogleScholarPublications(req, res) {
...
-  await cluster.close();
+  await browser.close();
 
-  console.timeEnd('doSomething');
+  await Promise.all(links.map((x) => scrapeGoogleScholar(x))).then((pub) =>
+    publications.push(...pub)
+  );
+
+  console.timeEnd('timeToScrape');
+  console.log(publications);
```
aksh0001, line 180: Would this log statement be too cloudy?

api/src/controllers/publications.js
```diff
@@ -132,9 +132,9 @@ function readAllPublicationsByTeam(req, res, next) {
 }
 
 /**
- * Given a google scholar user id, this function uses a headless browser via Puppeteer to scrape
+ * Given a google scholar user id, this function uses a headless browser via Playwright to scrape
  * the publications info from a user's profile. This runs several threads in parallel specified in the config.
```
aksh0001, line 136: If we're not running several threads, we should update this doc.

api/src/controllers/publications.js
```diff
@@ -154,7 +156,7 @@ async function getGoogleScholarPublications(req, res) {
     playwrightConfig.pageSizeSuffix +
     pageSize +
     playwrightConfig.sortBySuffix;
-  console.log(url);
+  logger.info(url);
```
aksh0001, line 159: Maybe add some more context to the log statement?
e.g. `GScholar Publications for user xxx = ${url}`

api/src/controllers/publications.js
```diff
@@ -192,8 +193,14 @@ async function getGoogleScholarPublications(req, res) {
...
 
+/***
+ * The function performs the scraping logic to get the Google Scholar publications
+ * and returns it in a format that fits the publication model.
+ * @see models/publication.model.js
+ * @returns a publication
+ */
 async function scrapeGoogleScholar(url) {
-  console.log(playwrightConfig.gScholarHome + url);
+  logger.info(playwrightConfig.gScholarHome + url);
```
aksh0001, line 203: Do we need both log statements (i.e.  the one on L159 and this one)?

aksh0001 (approved):
Nits only, but otherwise lgtm

api/src/controllers/publications.js
vgoh1: The one in line 203 is for the individual publication links while line 159 is for the profile link, which might be helpful for debugging the scraping logic for specific publications

api/src/config/playwright.js
```diff
@@ -2,10 +2,10 @@
  * This module stores the configuration for using Puppeteer to scrape Google Scholar.
  */
 
-const puppeteerConfig = {
+const playwrightConfig = {
   noOfDummyLinks: 4, // first 4 links aren't publications can be ignored
-  noOfThreads: 10, // number of concurrent threads
   pageSize: 10, // number of publications to return at once
+  gScholarHome: 'https://scholar.google.com.sg',
   baseUrl: 'https://scholar.google.com.sg/citations?hl=en&user=',
```
jfan0010, line 9: Just curious to know, why are we using Singapore URL instead of Australia?

jfan0010 (approved):
LGTM. Everything works expected on mine!

rpra0006 (approved):

api/src/config/playwright.js
rpra0006: Yes, was wondering about this too, scholar.google.com itself should be sufficient?

pmyip3 (approved):

api/src/config/playwright.js
vgoh1: Alright...By popular demand I will change it to just scholar.google.com :)

PR #139, opened 2021-07-31
Team profile redux state reset on refresh

pmyip3:
# Description 
There are 2 issues on the team profile page currently: 
1. the team information on the form will be cleared after refreshing the page 
2. after updating the team information and refresh the page, the profile will revert back to the previous team information 

# Type of change 
- [X] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
The problem with the above issue is that we had contains the teamName, orgName, and email in the JWT payload, and the way we send the team info from the backend to the frontend is by decoding the jwt payload. As a result, the team info will always show the old information unless we generate the new access token. 

# Solution description
Change the jwt payload to only contain the teamId, as it is not editable by users and then make request to the database to return the team info using the teamId
Also, use the useEffect hook to update the team info on the form whenever the relevant info is returned from the backend

# Tests 
All unit tests pass? N/A

Builds successfully? Yes 

N/A

The issue described in the description section should be fixed in this PR

# Risk
Low

### DIFF
--- a/api/src/controllers/team.js
+++ b/api/src/controllers/team.js
@@ -71,13 +71,19 @@ async function storeHandle(req, res, next) {
 }
 
 /**
- * Gets the team document from the auth middleware
- * @param {*} req request object, containing the team object 
- * @param {*} res response object, the team object 
- * @returns 200: return the team passed by the auth middleware 
+ * Gets the team info
+ * @param {*} req request object contains the teamId decoded in auth middleware
+ * @param {*} res response object, the team related info 
+ * @returns 200: the team related info  
  */
-function getTeam(req, res) {
-  return res.status(200).send(req.team);
+function getTeam(req, res, next) {
+  Team.findById(req.team._id).select('_id teamName orgName email')
+  .then((foundTeam) => {
+    if (foundTeam) {
+      return res.status(200).send(foundTeam);
+    }
+  })
+  .catch((err) => next(fillErrorObject(500, 'Server error', [err.errors])));
 }
 
 /**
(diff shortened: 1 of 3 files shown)


api/src/controllers/team.js
```diff
@@ -71,13 +71,19 @@ async function storeHandle(req, res, next) {
...
-function getTeam(req, res) {
-  return res.status(200).send(req.team);
+function getTeam(req, res, next) {
+  Team.findById(req.team._id).select('_id teamName orgName email')
+  .then((foundTeam) => {
+    if (foundTeam) {
+      return res.status(200).send(foundTeam);
+    }
+  })
+  .catch((err) => next(fillErrorObject(500, 'Server error', [err.errors])));
```
vgoh1, line 86: Because the get team logic has been moved into here, we might not need line 26 in middleware/team.js - because validating the teamId requires getting the team document anyway, getTeam and the middleware would be doing the same work. Up to you whether you want to change it or leave it as is.

vgoh1 (approved):

api/src/controllers/team.js
pmyip3: This get team endpoint requires the team_id, which is decoded from jwt in the previous auth middleware, not the team middleware. As this endpoint doesn't request any parameters sent from the frontend, only the accessToken is needed. 

While the validateTeamId middleware validates the teamId in the parameters sent from the frontend. 

They are doing different things though but will be a good idea to combine them together into only one middleware function, may be we can raise a ticket to investigate the way we can do that and refractor it.

vgoh1: Yeah, agree with combining them. Will put that on the board.

jfan0010 (approved):
Works fine on mine

PR #140, opened 2021-08-02
Display success message after successful action 

pmyip3:
# Description 
Refractor the current error handling components so that it could handle the success message sent from the backend as well

# Type of change 
- [ ] Bug fix
- [X] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
Some actions are hard to tell whether it is successful without showing a success message on the front end to the users. 
For example, after registration, the current website will redirect the user to the landing page without indicating whether the action is successful or not, which is not user-friendly. We should be able to display success messages to the user when necessary. 

# Solution description
The existing error handling components had been modified and extended so that it would be able to also handle success messages. 

Note: We might not want to display success messages when some actions are easy to tell whether it is successful (e.g. after login, after adding a publication). 

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

For now, only 2 places will pop up a success message after a successful action 
1. When register successfully 
image

2. When update team profile successfully 
image

- Try to register a team and update team info to see whether a success message will pop up.  

N/A

# Risk
Low-Medium might affect the error handling components


api/src/controllers/team.js
```diff
@@ -98,15 +98,17 @@ async function addTeam(req, res, next) {
       if (foundTeam) {
         return res.status(400).send('Email had been registered');
       }
+      bcrypt.genSalt().then((salt) => 
+        bcrypt.hash(req.body.password, salt).then((hashedPassword) => {
+          const hashedTeam = { ...req.body, password: hashedPassword };
+          Team.create(hashedTeam)
+            .then(() => {
+              return res.status(201).json('Team has been created successfully')
+            })
```
vgoh1, line 107: Do we need a catch here in case this fails to create the team?

vgoh1 (approved):
LGTM, would be great to see the success toast used in more places in the future to improve UX! Remember to update your branch with the latest main.

api/src/controllers/team.js
```diff
@@ -92,21 +92,23 @@ function getTeam(req, res, next) {
  * @param {*} res response object - updated team object
  * @returns 201: returns updated team details
  */
-async function addTeam(req, res, next) {
+function addTeam(req, res, next) {
```
aksh0001, line 95: This controller name doesn't seem too appropriate IMO. Add team implies a list of teams created in a particular session. This should be renamed to `createTeam`? @vgoh1 thoughts?

api/src/controllers/team.js
```diff
@@ -92,21 +92,23 @@ function getTeam(req, res, next) {
...
       if (foundTeam) {
         return res.status(400).send('Email had been registered');
       }
+      bcrypt.genSalt().then((salt) => 
+        bcrypt.hash(req.body.password, salt).then((hashedPassword) => {
+          const hashedTeam = { ...req.body, password: hashedPassword };
+          Team.create(hashedTeam)
+            .then(() => {
+              return res.status(201).json('Team has been created successfully')
+            })
```
aksh0001, line 107: curious: why are we going from try-catch style to promise style? Shouldn't the refactor be the other way around to avoid callback hell: https://blog.avenuecode.com/callback-hell-promises-and-async/await?
I'd suggest going back to simply async await, try-catch

api/src/controllers/team.js
```diff
@@ -92,21 +92,23 @@ function getTeam(req, res, next) {
...
+      );
     })
     .catch((err) => next(fillErrorObject(500, 'Server error', [err.errors])));
-
-  const salt = await bcrypt.genSalt();
-  const hashedPassword = await bcrypt.hash(req.body.password, salt);
-  const hashedTeam = { ...req.body, password: hashedPassword };
-  Team.create(hashedTeam)
-    .then((createdTeam) => res.status(201).json(createdTeam._id))
-    .catch((err) => next(fillErrorObject(500, 'Server error', [err.errors])));
```
aksh0001, line 109: Add a try-catch around this

client/src/reducers/mainReducer.js
```diff
@@ -1,19 +1,22 @@
...
+import { errorReducer } from '../notification/notificationReduxFunctions';
 
-const INITIAL_TEAM_STATE = {
+const INITIAL_NOTIFICATION_STATE = {
   error: null,
+  success: null,
 };
 
-const mainReducer = (state = INITIAL_TEAM_STATE, action) => {
+const mainReducer = (state = INITIAL_NOTIFICATION_STATE, action) => {
```
aksh0001, line 10: Please add docs here to explain what the `mainReducer` does?

client/src/notification/notificationReduxFunctions.js
```diff
@@ -1,6 +1,6 @@
 import { createSelector } from 'reselect';
```
aksh0001, line 1: File headers.

client/src/notification/notificationReduxFunctions.js
```diff
@@ -1,6 +1,6 @@
...
 import { get } from 'lodash';
-import { RESEARCHIFY_API_ERROR } from '../actions/types';
+import { RESEARCHIFY_API_ERROR, SUCCESS_MESSAGE } from '../actions/types';
```
aksh0001, line 3: Sorry I didn't know where to put this comment. But it seems that the `errorActionCreator` is not used anywhere - can this be deleted or do you see a use for it? IMO, if we don't have usages, it's best to remove it and when we do need it we can add it back in. 

See **YAGNI**: https://martinfowler.com/bliki/Yagni.html

client/src/notification/notificationReduxFunctions.js
```diff
@@ -44,6 +44,13 @@ export const errorActionGlobalCreator = (error) => {
...
 };
 
+export const successMessageCreator = (successMessage) => {
+  return {
+    type: SUCCESS_MESSAGE,
+    payload: successMessage
+  }
+}
+
 // FIXME: remove this if we don't have any selector
```
aksh0001, line 54: Can we remove `createErrorSelector`?

client/src/actions/team.js
```diff
@@ -16,24 +16,16 @@ import {
...
-    const teamData = {
-      ...teamInfo,
-      teamId: teamId,
-    };
-    // TODO: do we need to dispatch this action? 
-    dispatch({
-      type: ADD_TEAM,
-      payload: teamData,
-    });
+    const successMessage = await api.addTeam(teamInfo)
```
aksh0001, line 27: As with my comment on the API, does renaming this function to `createTeam` make it more meaningful? Also, the action creator should be renamed accordingly as well (i.e. `addTeamInfo`)

api/src/controllers/team.js
```diff
@@ -257,7 +259,7 @@ function updateTeam(req, res, next) {
     new: true,
     runValidators: true,
   })
-    .then((updatedTeam) => res.status(200).json(updatedTeam))
+    .then(() => res.status(200).json('Team has been updated'))
```
aksh0001, line 262: Why send a success message when we can just send back the `updatedTeam` object? This appears to break REST to me.

client/src/actions/team.js
```diff
@@ -287,11 +278,11 @@ function teamDataAllocator(teamData) {
  */
 export const updateTeam = (teamId, teamData) => async (dispatch) => {
   try {
-    const { data } = await api.updateTeam(teamId, teamData);
-    const updatedData = teamDataAllocator(data);
+    const successMessage = await api.updateTeam(teamId, teamData)
+    dispatch(successMessageCreator(successMessage));
```
aksh0001, line 282: Related to the above comment. We're dispatching an action with a payload that is set by the argument to this AC (i.e. `payload: {...teamData, teamId},`).

client/src/actions/team.js
```diff
@@ -287,11 +278,11 @@ function teamDataAllocator(teamData) {
  */
 export const updateTeam = (teamId, teamData) => async (dispatch) => {
   try {
-    const { data } = await api.updateTeam(teamId, teamData);
-    const updatedData = teamDataAllocator(data);
+    const successMessage = await api.updateTeam(teamId, teamData)
```
aksh0001, line 281: I see a potential design issue with requiring the api to determine the `successMessage`. Why place the burden on the API to determine what message to dispatch? Doesn't it make more sense for clients to examine the response from the API (which I believe we should go with the `updatedData` as before) and determine what message needs to be dispatched? This is a separation of concerns argument, and I think it'd be better for the client to determine what the message should be. 

I'd like more opinions on this though. @kdav108 @vgoh1

api/src/controllers/team.js
aksh0001: I did some research and it appears to be a contention topic: https://stackoverflow.com/a/798174/15507541
I'd still prefer returning the object over a success message though. A response code of `200` is perfectly enough to know that the team update operation was successful.

aksh0001: @vgoh1 @kdav108 @ssha0054

client/src/notification/notificationReduxFunctions.js
aksh0001: Also, does this have to be a component? Why not just a free function?

aksh0001 (changes requested):
I'm curious about the need to refactor the RESTful API design to support handling API successes in the client. I'd say the API response codes are enough to determine if an API request succeeded. If so, we can display the success toast.

api/src/controllers/team.js
pmyip3: Thanks for pointing out the issue. Yeah, I was struggling whether which way we should adopt too. 
I decided to send back a message instead of an object eventually cause I thought it would be easier to let all the success/ error messages be handled in the backend and I didn't find a strong reason (other than would be helpful for testing in postman) to send back the object as it's just the same object sent from the frontend(the updateTeam endpoint) or the frontend not require the object to be sent back (the createTeam endpoint).

client/src/notification/notificationReduxFunctions.js
pmyip3: Sure, will do

api/src/controllers/team.js
pmyip3: Sure, you reckon we raise a PR to convert all the promises back to async/await?

api/src/controllers/team.js
pmyip3: The winder-break team has created an error handler function, so any error the function cannot be handled will be passed to the error handler. So I think try catch blocks will no longer need. The error handler function is in api/server.js currently. 
A bit of info: https://expressjs.com/en/guide/error-handling.html

api/src/controllers/team.js
vgoh1: To get technical since it is a patch operation, [convention](https://stackoverflow.com/questions/54666244/what-should-response-of-rest-full-api-for-patch-http-method) says returning `204 No Content` is possible if the client knows what the updated resource is going to be (which iirc they will cause this is called in the team profile page on the frontend). But to address the question of whether it should be returning the object or not, then yes it should to follow convention or it can return 204. On a related note: I agree with Akshay's comment [below](https://github.com/Researchify/Researchify/pull/140#discussion_r681403135) for reasons that I will post in that thread...

client/src/actions/team.js
vgoh1: I agree, the client should determine the message to show the user rather than depending on the backend because it's easier to check the HTTP code of the API call rather than message for maintainability reasons such as:
- not having to go through to the backend to deal with the success/error handling for the frontend if we have to make changes later,
- HTTP codes are a reliable way to tell whether the operation was successful or not which the frontend can use. And because we know what endpoint it's calling the success message can be handled by the frontend.

api/src/controllers/team.js
aksh0001: Yep, that'd be great!

api/src/controllers/team.js
aksh0001: Makes sense. Resolving.

api/src/controllers/team.js
aksh0001: Yes, I think one of the strong arguments to revert back to returning an object is discussed in the comment thread below. Basically:
- I think we don't need to return a success **message**, because the response code is sufficient
- As to Mable's point about finding a strong reason to send back the same object in update operations: I don't have a strong reason either, other than the fact that it makes the API cleaner, see this comment: https://github.com/Researchify/Researchify/pull/140#discussion_r681402217

api/src/controllers/team.js
```diff
@@ -247,22 +244,22 @@ async function deployToGHPages(req, res, next) {
  * @returns 404: team is not found
  * @returns 400: team id is not in a valid hexadecimal format
  */
-function updateTeam(req, res, next) {
+// eslint-disable-next-line no-unused-vars
```
aksh0001, line 247: nit:this is my personal preference, but consider inlining this eslint message to the same line as the function definition (L248). See: https://github.com/Researchify/Researchify/blob/main/api/src/middleware/error.js#L37

api/src/controllers/team.js
```diff
@@ -247,22 +244,22 @@ async function deployToGHPages(req, res, next) {
...
   const team = req.body;
 
-  Team.findByIdAndUpdate(_id, team, {
+  const updatedTeam = await Team.findByIdAndUpdate(_id, team, {
     new: true,
     runValidators: true,
   })
-    .then((updatedTeam) => res.status(200).json(updatedTeam))
-    .catch((err) => next(fillErrorObject(500, 'Server error', [err])));
+  res.status(200).json(updatedTeam)
```
aksh0001, line 256: Is there a possibility this will throw if the update was unsuccessful, in which case we should wrap with a `try-catch` and if an error occurs, fill the error object?

aksh0001 (approved):
LGTM. Nice work! please see my two (minor) comments on eslint warnings and about handling update team errors.

api/src/controllers/team.js
pmyip3: I think it is what the error middleware is responsible for. If findByIdAndUpdate throws an error(which I think the error might be server error or anything else we could not handle, as the teamId is already been verified by the validateTeamId middleware before this function), next will be called with a default Error object provided by the Express router. In this case, a default error object (no the type of fillErrorObject we created) will be passed into the error middleware, and the error middleware will handle it.

vgoh1 (approved):

vgoh1:
Merging prematurely because I need this for my feature.

PR #142, opened 2021-08-05
Improve import publications logic

vgoh1:
# Description 
The current logic does the following: it retrieves the first 10 publications based on publication date (latest publications first), and if all 10 are already in the database, then it say 'All publications from the profile have already been imported', which isn't foolproof

This needs to be improved to be more thorough ie just let it go through the entire profile before returning it, however it needs to differentiate between publications already being in the db vs no publications from the scraper (end of profile)

# Type of change 
- [x] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
See description

# Solution description
Backend:
So if you reach the [end of the profile](https://scholar.google.com.sg/citations?user=3tyGlPsAAAAJ&hl=en&cstart=50&pagesize=10), there aren't any publication links you can click on. Knowing that there are always 4 clickable links on the page (noOfDummyLinks), if the scraper doesn't get more than 4 links, then we've reached the end of the profile. There's a flag in the response (reachedEnd) that will tell the frontend whether it's reached the end of the profile.

Frontend:
The frontend uses this flag to differentiate between pubs being in the db vs reaching end of the profile. It will show a success toast for every scrape it does that doesn't return any publications. The imported pubs page has also been changed to show a message if no publications were retrieved. Also refactored the importPublication action into one function and pulled the validation up.

# Tests 
npm run dev in /api, npm run start in /client
Click on import on the publications page, enter profile url

I've tested the edge cases I could think of:
- All publications of the user has been imported -> no publications till the end
- Only the first 10 have not been imported -> the first 10 pubs should show until the end
- The second set of 10 publications have not been imported -> similar result to before

Please run it so the testing can be more extensive!

N/A

# Risk
Low, should be isolated to the import functionality


client/src/components/publications/form/ProfileLinkPage.js
```diff
@@ -18,9 +20,29 @@ const ProfileLinkPage = ({ closeModal }) => {
...
+    } else {
+      gScholarId = values.profileLink.substring(position + 5, position + 17);
+      dispatch({
+        type: UPDATE_GSCHOLAR_ID,
+        payload: gScholarId,
+      });
+      return true;
+    }
+  };
+
```
vgoh1, line 41: Wasn't sure how else to pull up the validation from the actions, please let me know if it can be better. This was pulled up after discussion with @pmyip3 since the validation logic didn't need to be in the action.

client/src/notification/NotificationToaster.js
```diff
@@ -5,21 +5,23 @@
...
+    if (success) {
       toast.success(success);
     }
-  }, [success])
-  
+    dispatch({
+      type: CLEAR_NOTIFICATION,
+    });
+  }, [error, success]); // eslint-disable-line react-hooks/exhaustive-deps
+
```
vgoh1, line 25: This was added by @pmyip3 after we found out that the toast didn't show up if the message was the same due to useEffect()

api/src/controllers/publications.js
```diff
@@ -168,27 +170,46 @@ async function getGoogleScholarPublications(req, res) {
...
-  console.timeEnd('timeToScrape');
+    await Promise.all(links.map((x) => scrapeGoogleScholar(x))).then((pub) =>
+      publications.push(...pub)
+    );
 
-  const newPublications = await validateImportedPublications(
-    teamId,
-    publications
-  );
+    console.timeEnd('timeToScrape');
```
aksh0001, line 194: Do we need to log timing info in prod? I'd suggest maybe enabling this only for dev/debug mode (e.g. through the use of environment variables, and if, for example: `process.env.DEBUG` evaluates to true, we can enable these log statements.

api/src/controllers/publications.js
```diff
@@ -168,27 +170,46 @@ async function getGoogleScholarPublications(req, res) {
...
-    retrieved: publications.length,
-    newPublications: newPublications,
-  };
+    const newPublications = await validateImportedPublications(
+      teamId,
+      publications
+    );
+
+    // if the no of pubs is less than the page size,
+    // then we've reached the end of the profile
```
aksh0001, line 202: nit: prefer full sentences w/ the comments and spell out everything

api/src/controllers/publications.js
```diff
@@ -158,6 +158,8 @@ async function getGoogleScholarPublications(req, res) {
     playwrightConfig.sortBySuffix;
   logger.info(`GScholar profile for user id ${author}: ${url}`);
   let publications = [];
+  let endOfProfile = false;
+  let response;
```
aksh0001, line 162: Consider initializing the  `response` here, with default values:
```
   response = {
      retrieved: publications.length,
      newPublications: [],
      reachedEnd: endOfProfile,
    };
    ```
    
to avoid some duplication below

File: api/src/controllers/publications.js [hunk 6]
```diff
@@ -168,27 +170,46 @@ async function getGoogleScholarPublications(req, res) {
...
+    // no pubs found
+    endOfProfile = true;
+    await browser.close();
 
-  await browser.close();
+    response = {
+      retrieved: publications.length,
+      newPublications: [],
+      reachedEnd: endOfProfile,
+    };
```
aksh0001, line 182: get rid of this if we initialize `response` above

api/src/controllers/publications.js
```diff
@@ -168,27 +170,46 @@ async function getGoogleScholarPublications(req, res) {
...
+    }
 
-  await Promise.all(links.map((x) => scrapeGoogleScholar(x))).then((pub) =>
-    publications.push(...pub)
-  );
+    await browser.close();
 
-  console.timeEnd('timeToScrape');
+    await Promise.all(links.map((x) => scrapeGoogleScholar(x))).then((pub) =>
+      publications.push(...pub)
```
aksh0001, line 191: end statements with `;`?

api/src/controllers/publications.js
```diff
@@ -168,27 +170,46 @@ async function getGoogleScholarPublications(req, res) {
...
+    // then we've reached the end of the profile
+    if (publications.length < pageSize) {
+      endOfProfile = true;
+    }
+
+    response = {
+      retrieved: publications.length,
+      newPublications: newPublications,
+      reachedEnd: endOfProfile,
+    };
```
aksh0001, line 211: This may need tuning if initializing `response` above

api/src/controllers/publications.js
```diff
@@ -225,7 +246,7 @@ async function scrapeGoogleScholar(url) {
   const publicationInfo = {};
   fields.forEach((key, i) => (publicationInfo[key] = values[i]));
 
-  // TODO: this logic depends on the order of the fields, 
+  // TODO: this logic depends on the order of the fields,
   // which will differ based on the info of the publication, can be improved
```
aksh0001, line 250: Can you be more descriptive with what specifically needs to be done in this `TODO`? In case someone else wants to pick it up

client/src/actions/publications.js
```diff
@@ -109,82 +112,54 @@ export const sortPublications =
...
-      if (result.data.retrieved < pageSize) {
-        // reached the end of the user's publications
         dispatch({
           type: IMPORT_END,
         });
+      } else if (
+        result.data.newPublications.length === 0 &&
+        result.data.retrieved > 0
+      ) {
+        // no new pubs retrieved but not end of profile
```
aksh0001, line 145: nit: prefer spelling out `pubs`

client/src/actions/publications.js
```diff
@@ -109,82 +112,54 @@ export const sortPublications =
...
+        dispatch({
+          type: IMPORT_EMPTY,
+        });
+        dispatch(
+          successMessageCreator(
+            'No new publications were found on page ' + pageNo
```
aksh0001, line 151: consider using string templates. Ditto for other usages.

client/src/notification/NotificationToaster.js
```diff
@@ -5,21 +5,23 @@
...
   useEffect(() => {
-    if(error){
+    if (error) {
       toast.error(error.data);
     }
-  }, [error])
-
-  useEffect(() => {
-    if(success){
+    if (success) {
```
aksh0001, line 18: Could there  be a scenario where both these evaluate to true and both success + error toasts show up? I'm thinking if this is possible because of the fact that we are storing both error and success notifications in our store.

api/src/controllers/publications.js
vgoh1: After your comment, I'm thinking I'll probably get rid of them as a whole because playwright is looking pretty consistent and stable right now and (fingers crossed) it won't break anytime soon.

api/src/controllers/publications.js
vgoh1: If you're talking about line 191, then it doesn't need a `;` there cause it causes a syntax error, plus prettier doesn't put one there either.

api/src/controllers/publications.js
vgoh1: It's a bit complicated so I'm going to explain it in a ticket instead, and it's probably not an issue until someone reports it as a bug with specific publications.

client/src/notification/NotificationToaster.js
vgoh1: The only time I can think of it happening is if someone intentionally uses both error and success toasts at the same time, but I'll change it into an else if so that it'll default to the error toast instead of both in that case.
@pmyip3 Feel free to add if you have any thoughts.

api/src/controllers/publications.js
aksh0001: Sorry, I meant that `push` invocation on the list

client/src/notification/NotificationToaster.js
pmyip3: I agree that it will work fine other than someone intentionally uses both error and success toasts at the same time, as the notification state would be cleared after each rendering of the NotificationToaster because we dispatch a CLEAR_NOTIFICATION action inside the useEffect hook. 
Not quite sure what would happen if we dispatch both error and success message actions at the same time. Cause I don't think it will wait for the completion of the first dispatched action ( as it will go through the state, then the toastNoticaition would get updated because of the changes in state). So we might get both success and error messages existing in the reducer in this situation.
 Another possible solution is we can explicitly clear the error messages in the state when dispatching success messages, vice versa. 
`    case SUCCESS_MESSAGE:
      return {...state, error: null, success: action.payload }`
Then we can ensure only one of the messages would exist in the state at any point of time.

api/src/controllers/publications.js
aksh0001: SG

api/src/controllers/publications.js
aksh0001: SG

api/src/controllers/publications.js
aksh0001: Woops, I misread the code. You're right. Resolving this comment.

aksh0001 (approved):
LGTM. Will defer to Mable for handling the success/error notifications bug

pmyip3 (approved):

cyr17 (approved):
This is a good way to prevent huge waiting times for importing publications

jfan0010 (approved):
LGTM! Works smoothly.

After importing the first few pages (eg. first 4 pages), the user needs to click the `show more` button multiple times (depends on how many pages you've imported) in order to get the next publication that hasn't been imported. But functionality wise, its great!

PR #146, opened 2021-08-07
Removed files for editor site.

jfan0010:
# Description 
Editor page is no longer needed in Researchify. Files for editor page to be removed and cleaned up.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Tests 
All unit tests pass? 
N/A
Builds successfully?
Yes

# Risk
Small. Might have missed some files.


vgoh1 (approved):

client/src/components/route/PrivateRoute.js
```diff
@@ -47,8 +45,6 @@ const PrivateRoute = () => {
           />
           <DashboardLayoutRoute path="/team" exact component={TeamPage} />
```
pmyip3, line 46: Since we get rid of the EditorLayoutRoute component, can we also get rid of the DashboardLayoutRoute component and simply use <Route path="/team" exact component={TeamPage} /> like what we have in the PublicRoute component?

jfan0010: I would say keeping `DashboardLayoutRoute`, because it actually has the `header` and `sidebar` in it, if we take the page content out, we'll have to put `header` and `sidebar` in `PrivateRoute.js`, which I personally think that is a bit messy when looking into `PrivateRoute.js`.  What do you think?

pmyip3: True. Make sense to me.

pmyip3 (approved):

rpra0006 (approved):

cyr17 (approved):
NIce

PR #148, opened 2021-08-09
Add fake data to run client website locally

kdav108:
# Description 
Adds fake data for the client website so we can run it as a standalone React app to debug.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
Currently, if we run `nom run start` from `schools/base/` it fails because the environment variables it depends on to get data about a team's publications and team info are not set. 

# Solution description
This PR adds fake data so if we're running the React App by itself, it uses the fake data instead of the environment variables.

# Tests 
All unit tests pass?
Yes
Builds successfully?
Yes

1) Go into `schools/base/` and run `npm run start`
2) Hopefully you'll be able to see the client website at localhost:3000

# Risk
High, if the logic inside `scholly/base/src/global/data.js` is incorrect, when deploying a client website we may accidentally publish the fake data instead of the correct data from the environment variables.


kdav108:
FYI I haven't actually done any regression testing to see if deployment still works with the correct client data. I'm just struggling to get deployment working on my computer so if someone can test that during PR review that'd be great!

scholly/base/src/global/data.js
```diff
@@ -6,13 +6,38 @@
...
+} from './fakeData';
 
+const env = process.env;
+let TEAM_PUBLICATIONS;
+let TEAM_INFO;
+let TEAM_MEMBERS;
 
-/// The list of publications the team has created for rendering in the publications page
-export const TEAM_PUBLICATIONS = JSON.parse(process.env.REACT_APP_TEAM_PUBLICATIONS);
+if (env.REACT_APP_DEPLOYING === 'true') {
```
aksh0001, line 20: Prefer calling this environment variable `REACT_APP_DEBUG` so that we can default to fake data if we're running in development mode.

aksh0001: Also, we should document this somewhere: i.e. that it is possible to run in Debug mode using this env var, which will run the base app using fake data. Do you reckon writing this process up in a gdocs?

aksh0001: Also, add a comment/jsdoc somewhere (e.g. above this line or in the file header) noting that we use fake data if this env var is set.

aksh0001: Also, prefer not to do a string comparison for the environment variable. What about just checking if `REACT_APP_DEBUG` is set? https://stackoverflow.com/questions/30047205/how-can-i-check-if-an-environment-variable-is-set-in-node-js

scholly/service/core/build.js
```diff
@@ -38,6 +38,7 @@ async function buildBaseApp(data) {
         [REACT_APP_TEAM_PUBLICATIONS]: JSON.stringify(data.teamPublications),
         [REACT_APP_TEAM_INFO]: JSON.stringify(data.teamInfo),
         [REACT_APP_TEAM_MEMBERS]: JSON.stringify(data.teamMembers),
+        [REACT_APP_DEPLOYING]: 'true',
```
aksh0001, line 41: Do we need this? Why expose the env var to the scholly service if it's just used for debugging. We could just set the environment variable manually and run the base app w/out going through Scholly, if its for debugging purposes.

scholly/service/core/data.js
```diff
@@ -10,4 +10,5 @@ module.exports = {
   REACT_APP_TEAM_PUBLICATIONS: 'REACT_APP_TEAM_PUBLICATIONS',
   REACT_APP_TEAM_INFO: 'REACT_APP_TEAM_INFO',
   REACT_APP_TEAM_MEMBERS: 'REACT_APP_TEAM_MEMBERS',
+  REACT_APP_DEPLOYING: 'REACT_APP_DEPLOYING',
```
aksh0001, line 13: As above.

scholly/base/src/global/fakeData.js
```diff
@@ -0,0 +1,31 @@
...
+      categoryTitle: 'Example Journal',
+      issue: '1.1',
+      volume: '0.2',
+      pages: '20',
+      publisher: 'Penguine books',
+    },
+    link: 'thislinkisfakedonteventrytoclickit.okay',
+    authors: ['KK', 'LOL'],
+    yearPublished: '2007',
+  },
```
aksh0001, line 15: Please add > 1 publication to see the effects of having more publications that can fit in one page, hence needing scrolling.

scholly/base/src/global/fakeData.js
```diff
@@ -0,0 +1,31 @@
...
+  teamName: 'MonTeam',
+};
+
+export const FAKE_TEAM_MEMBERS = [
+  {
+    fullName: 'John',
+    position: 'Chief Scientist',
+    summary:
+      'John is a chief scientist at MonTeam, working with the top government agencies to fight the pressing issues arising from climate change',
+  },
```
aksh0001, line 30: Add more team members.

scholly/base/src/global/data.js
```diff
@@ -6,13 +6,38 @@
...
+  TEAM_PUBLICATIONS = env.REACT_APP_TEAM_PUBLICATIONS
```
aksh0001, line 21: Please add back the comment explaining what `TEAM_PUBLICATIONS` is for.

scholly/service/core/build.js
aksh0001: For linux/mac you could export:
`export REACT_APP_DEBUG=true && npm start`

aksh0001 (changes requested):

aksh0001:
This is a great idea btw, thanks for spearheading this! :)

scholly/service/core/build.js
aksh0001: So to clarify:
IMO we should just let the base react app know about the debug environment variable so we won't have to make changes to the scholly service. This can be achieved by setting the env var manually before running the base app.

scholly/base/src/global/fakeData.js
kdav108: Yep I was just a bit too lazy to make more fake data 😂

scholly/base/package.json
```diff
@@ -17,7 +17,7 @@
     "web-vitals": "^1.1.2"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "cross-env REACT_APP_DEBUG=true react-scripts start",
```
aksh0001, line 20: Great idea for that `cross-env` package, I didn't know about that! That should help w/ env var compatibility across platforms.
I would strongly suggest placing the command `"cross-env REACT_APP_DEBUG=true react-scripts start"` under a `dev` script so that the deployment in prod still uses the real data - cuz otherwise the fake data will be used there as well.

E.g.
```
 "scripts": {
    "dev": "cross-env REACT_APP_DEBUG=true react-scripts start",
    "start": "<leave old start script>"
...
}
```

So now whoever wants to run the app in debug mode can just execute `npm run dev`.

aksh0001 (approved):
LGTM. Please place the new command under a `dev` npm script and retain the old one so that we won't cause regressions in prod.

rpra0006 (approved):

cyr17 (approved):
Looks good,

Brainstorming an idea right now for onboarding, do you reckon we should keep this dummy data for the onboarding process so that the user can see what happens when they click deploy (Or is this a bad idea).

scholly/base/package.json
kdav108: Ahh right. Just wondering though, in the scholly service we invoke `npm build` in the base app, does that then run `npm start`?? If it doesn't then it shouldn't matter much if we add the cross-env stuff in dev or start script, I'm guessing

kdav108:
> Looks good,
> 
> Brainstorming an idea right now for onboarding, do you reckon we should keep this dummy data for the onboarding process so that the user can see what happens when they click deploy (Or is this a bad idea).

If it helps, why not. But I guess for on boarding if we're going to have a guide then we just show photos of what the deployed site looks like

PR #149, opened 2021-08-10
Fix import publications pagination

vgoh1:
# Description 
Few issues were found with the pagination after importing publications:
1. Clicking on the page numbers in a non-consecutive order would cause the visible publications to disappear, and it usually happens when you go from a larger page number to a smaller one
2. The logic for changing the page the user is on after an import was a bit weird in that if you weren't on the latest page, it would just increment your current page by 1.
Also did some general refactoring for the importPublicationsReducer and added a small fix for the scraping logic in the backend.

# Type of change 
- [x] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
See description

# Solution description
1. The logic for calculating the publications to show was wrong so this has been fixed in L151-152 in ImportSuccessPage.
2. In importPublicationsReducer.js, I decided to simplify the code to handle 2 cases, see L30-36.

# Tests 
Run api and client, and import publications and click around the pages to see if there's some weird behaviour (there shouldn't be!)

N/A

# Risk
Low, isolated to the import functionality


api/src/controllers/publications.js
```diff
@@ -221,10 +222,14 @@ async function scrapeGoogleScholar(url) {
...
   await page.goto(playwrightConfig.gScholarHome + url);
-
-  const title = await page.$$eval('a.gsc_oci_title_link', (titles) =>
+  let title = await page.$$eval('a.gsc_oci_title_link', (titles) =>
     titles.map((title) => title.innerText)
-  );
+  ); // we assume the publication title is a link
+
+  if (title[0] === undefined) { // if its undefined, then it wasn't a link
+    title = await page.$$eval('div[id=gsc_oci_title]',(titles) =>
```
vgoh1, line 230: See example of publication with no link here: https://scholar.google.com/citations?view_op=view_citation&hl=en&user=bbEQGY8AAAAJ&sortby=pubdate&citation_for_view=bbEQGY8AAAAJ:GbL5_m8SeH4C

rpra0006 (approved):
Looks good

api/src/controllers/publications.js
```diff
@@ -221,10 +222,14 @@ async function scrapeGoogleScholar(url) {
...
+        titles.map((title) => title.innerText))
```
aksh0001, line 231: End line with `;`?

aksh0001 (approved):

pmyip3 (approved):

PR #152, opened 2021-08-10
Sign up validation

pmyip3:
# Description 
Add validation for teamName, orgName, email, password, and confirmed password in frontend before calling the createTeam API endpoint

# Type of change 
- [ ] Bug fix
- [X] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
Currently, the program does not validate teamName, OrgName and the length of the passwords in the frontend 

# Solution description
Refractor code using formik with validations which can be found in the teamInfoSchema in client/src/components/auth/RegistrationForm.js 

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

image
image

# Risk
Low, only the register component and createTeam endpoint might be affected if bugs occurs


client/src/components/auth/RegistrationForm.js
```diff
@@ -11,139 +11,153 @@ import { createTeam } from '../../actions/team';
...
+      .required('Team Name is required')
+      .min(3, 'Must be at least 3 characters'),
+    orgName: yup
+      .string()
+      .required('Organization Name is required')
+      .min(3, 'Must be at least 3 characters'),
+    email: yup
+      .string()
+      .email('Invalid Email')
+      .required('Email is required'),
```
vgoh1, line 40: Can we improve this to have proper email domain validation? Not sure how complicated it'll get though

client/src/components/auth/RegistrationForm.js
```diff
@@ -11,139 +11,153 @@ import { createTeam } from '../../actions/team';
...
+    password: yup
+      .string()
+      .required('Please Enter your password')
+      .matches(
+        /^(?=.*[A-Za-z])(?=.*\d)(?=.*[@$!%*#?&])[A-Za-z\d@$!%*#?&]{8,}$/,
+        "Use 8 or more characters with a mix of letters, numbers & symbols"),
```
vgoh1, line 46: For people who can't decode regex at a glance, what's the pattern here in plain words?

client/src/components/auth/RegistrationForm.js
aksh0001: If we can have password regex, we should be able to do it for email as well. The regex for email is widely documented online, e.g. on StackOverflow

aksh0001 (approved):

client/src/components/auth/RegistrationForm.js
rpra0006: Maybe have a check to see if '@' is at least present, as a valid email.

rpra0006 (approved):

client/src/components/auth/RegistrationForm.js
pmyip3: I think the built-in .email() function from formik already checking if '@' is present when validating the email.
In terms of checking the proper domain, I had a search but couldn't find any relevant regex. I don't think it's possible to do it cause a new email domain can be registered by anyone unless we want to be restricted the input email to a specific domain e.g. google.com.

client/src/components/auth/RegistrationForm.js
vgoh1, line 46: This message can be improved to be more specific about the password policy, that you need at least 1 digit and symbol on top of the characters.

vgoh1 (approved):

PR #156, opened 2021-08-12
Add more details in publication fake data

jfan0010:
So I've added some details into the publication fake data, to look more 'real'.
Things I've added:
- Longer description
- _id
- title of the publication
- changing the category type to all caps, cause that's what is stored I believe. And changed one of them to 'OTHER'


vgoh1:
Actually, you could've probably just done this in his branch - feel free to merge.

PR #163, opened 2021-08-16
Integrate Scholly with AWS Deployment

aksh0001:
# Description 

This PR aims to solve a couple of things, but the goal is to integrate Scholly with our AWS workflow. As such this PR:
- Updates Dockerfiles (and development Dockerfiles), docker-compose file (used for local testing), and Dockerrun.aws.json (used in prod) to integrate Scholly
- Updates configuration in the `client` and `api` to use the new GitHub OAuth credentials from the Researchify organization account
- Includes logic to decide which OAuth credential set to use, depending if we're running locally/development or in production using the conventional `NODE_ENV` environment variable
- Removes no-longer-used dependencies (found using the `depcheck` utility) to reduce the built image sizes
- Uses a new library on the `client` called `react-github-login` to handle the initial OAuth on the client side to fetch the temporary code to be used for exchange with an OAuth access token 
- Updates CI configuration to build and push the new Scholly image

# Type of change 
- [x] Bug fix
- [x] New feature
- [x] Refactoring
- [x] Documentation 

# Problem

Scholly is a new addition to Researchify that builds and deploys a client website to GitHub pages, and as such it needs to be integrated with our development and production flows.

# Solution description

See Description.

# Tests 
All unit tests pass? Y

Builds successfully? Y

Watch: https://drive.google.com/file/d/1yo1saXmofxZ5wvBA5XVxHirBagr3GKe7/view

# Risk
Very High. These changes affect the full E2E flow both in development and production.


scholly/Dockerfile
```diff
@@ -1,19 +1,19 @@
-FROM node:alpine
+FROM node
 
+USER node
+
+RUN mkdir -p '/home/node/app'
```
kdav108, line 5: Why do we need to make a directory here? We don't seem to do it in any of the other Dockerfiles (for `client` and `api`)

docker-compose.yml
```diff
@@ -19,6 +19,14 @@ services:
       - /home/node/app/node_modules
       - ./api:/home/node/app
 
+  scholly:
```
kdav108, line 22: Not related to the work on this PR, but can we map port 3000 in client container to 3000 on the host, I think it would be easier to just have the same link to the website regardless of whether you're running locally or in docker. 

Plus, for some reason the website doesn't load (times out) if I use the other address assigned by docker and not the usual `http://localhost:3000`

kdav108 (changes requested):
Thanks for fixing/cleaning/adding these things! Just got a few suggestions and questions (out of my own curiosity)...

Also, if it's feasible it'd be great if we can split the PR in the future (if there's a lot of changes being made). Looking at this figure just makes me want to procrastinate sometimes 😂😂, although most of these files are just small changes or just package.json files
Screen Shot 2021-08-17 at 9 50 47 pm

kdav108:
Forgot to mention 2 more things:

1) Probably for another PR, but just writing it here for the record, the current deploy button is not very intuitive as a button (or maybe it's just me). It'd be great if we can wrap the functionality around another React button and hide this original one. Let me know what you guys think.

Screen Shot 2021-08-17 at 9 28 13 pm

2) Didn't get much time digging into this, but I get this error when running from docker as soon as the page loads (and I'm auto-signed in). I remember getting this error before and I think we had a fix for it?? Just can't remember what it was
Screen Shot 2021-08-17 at 9 27 07 pm

vgoh1:
Agree with Kunj's comment about not liking the deploy button, I preferred the old one.
A question re the time it takes for scholly to do its thing: It feels way longer than how long it takes locally, is there a reason why? Is it possible to speed it up? What happens if the user clicks to another page while scholly is still working in the background?

api/src/config/deploy.js
```diff
@@ -1,14 +1,28 @@
...
 
-const githubClientId = '03757c278dedad4b307d';
+/// Hostname of the scholly service.
+let schollyHost;
+/// Client ID of the configured GitHub OAuth application.
+let githubClientId;
+
+// Values depend on whether we're running in dev mode or prod (in containers).
+if (process.env.NODE_ENV === 'production') {
+  schollyHost = 'http://scholly:8000';
```
vgoh1, line 14: Is this just referring to the container name like we saw in your demo video?

client/src/components/dashboard/deploy/DeployPage.js
```diff
@@ -18,31 +20,39 @@ const DeployPage = ({ teamId }) => {
...
 
   const GitHubLoginButton = (
-    <Button
-      className="float-right"
-      variant="outline-primary"
-      href={githubLoginUrl}
-      disabled={retrievedAccessToken}
-    >
+    <GitHubLogin className="float-right" clientId={githubClientId} scope={scope}
+                 onSuccess={onSuccessfulLogin} onFailure={onLoginFail}
```
vgoh1, line 52: Is the reason why you decided to use this GH login component over the previous one is because it has support for checking the login result? That is neat that we don't have to rely on useEffect for this.

client/src/reducers/teamReducer.js
```diff
@@ -21,6 +21,7 @@ const INITIAL_TEAM_STATE = {
   themeId: '',
   repoCreated: false,
   error: null,
+  retrievedAccessToken: false,
```
vgoh1, line 24: I think this variable can straight up be removed because the only place it was being used was in DeployPage.js

Dockerrun.aws.json
```diff
@@ -6,14 +6,24 @@
...
     },
     {
       "name": "api",
       "image": "ak1738/researchify-api",
       "hostname": "api",
       "essential": false,
-      "memory": 128
+      "memory": 256,
+      "links": [
+        "scholly"
```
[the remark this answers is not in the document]
aksh0001: Good catch. I belive `links` is required for the `Dockerrun.aws.json`: https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/create_deploy_docker_v2config.html

But for `docker-compose`, all containers can discover each other by default so `links` isn't required per se.

Dockerrun.aws.json
```diff
@@ -6,14 +6,24 @@
...
+      ]
+    },
+    {
+      "name": "scholly",
+      "image": "ak1738/researchify-scholly",
```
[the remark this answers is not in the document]
aksh0001: Yep. I had a plan to create a Researchify Docker Hub account last sem. But this will take some setup and rn it's not a priority.

api/src/controllers/team.js
```diff
@@ -187,42 +185,76 @@ function updateTeamMember(req, res, next) {
...
     );
   }
 
-  const ghUsername = response.data.login;
-  console.log(ghUsername);
+  const ghUsername = data.login;
+  logger.info(`GitHub deploy initiated for user: ${ghUsername}`);
 
+  // TODO: ideally this data should be fetched by us, and we should not expect
+  //  the client to provide it.
```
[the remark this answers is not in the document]
aksh0001: I think it's best that the data that needs to be sent to Scholly should be fetched by the `API` - atleast that was how I originally designed it. This is a separation of concerns argument and also the fact that it'll make the entire [CUJ](https://www.reforge.com/brief/critical-user-journeys-how-google-product-teams-react-when-growth-slows#2SsLPAALDeKxpwQPdEp6Uw) when they deploy the app slower since if they have 100 publications all of that is sent over the wire, and unnecessarily. I prefer the old design where the API is responsible for fetching all this info to feed into scholly.

client/src/components/dashboard/deploy/DeployPage.js
```diff
@@ -18,31 +20,39 @@ const DeployPage = ({ teamId }) => {
...
+  //   if (hasCode && !retrievedAccessToken && teamId) {
+  //     const code = url.split('?code=')[1];
+  //     // we use this code to exchange an access token
+  //     dispatch(getGHAccessToken(teamId, code));
+  //   } else if (!retrievedAccessToken) {
+  //     // we refreshed so we should clear local storage
+  //     localStorage.clear();
+  //   }
+  //   // eslint-disable-next-line react-hooks/exhaustive-deps
+  // }, [dispatch, teamId]);
```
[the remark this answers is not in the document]
aksh0001: So related to your comment about button aesthetics, I'll need someone to work on that as a follow up to this PR, and they can refactor away this code too.

scholly/Dockerfile
aksh0001: Great question, so:
1. Containers by default execute as `root` user
2. The `WORKDIR` command creates non-existing directories and sets the owner to `root`
3. The `USER` directive specifies that the container should run as `node`
4. Scholly creates that `build` folder inside this directory when building an application
5. `node` user does not have more permissions than `root` so we have a permission conflict
6. This change makes it so that `node` owns this directory

Why do we want to execute as `node` user and not `root`? It's about security: see point 2 in https://snyk.io/blog/10-docker-image-security-best-practices/

https://docs.docker.com/engine/reference/builder/#user

Ref: https://github.com/nodejs/docker-node/issues/740

docker-compose.yml
aksh0001: You should use the nginx proxy locally which exposes `3050`

aksh0001:
> Thanks for fixing/cleaning/adding these things! Just got a few suggestions and questions (out of my own curiosity)...
> 
> Also, if it's feasible it'd be great if we can split the PR in the future (if there's a lot of changes being made). Looking at this figure just makes me want to procrastinate sometimes , although most of these files are just small changes or just package.json files
> Screen Shot 2021-08-17 at 9 50 47 pm



> Agree with Kunj's comment about not liking the deploy button, I preferred the old one.
> A question re the time it takes for scholly to do its thing: It feels way longer than how long it takes locally, is there a reason why? Is it possible to speed it up? What happens if the user clicks to another page while scholly is still working in the background?

AFAIK there is no way to speed this up. We have 4 containers running on a 1gb ram `t2 micro`. Re: button, I agree. I've addressed this in Kunj's comment above.

client/src/reducers/teamReducer.js
aksh0001: Added because we're referencing it in L53: https://github.com/Researchify/Researchify/blob/fd5babb72485a8c34041f8681e25af4e042eec1e/client/src/reducers/teamReducer.js#L53

client/src/components/dashboard/deploy/DeployPage.js
aksh0001: Yes! I found it much cleaner, but the downside is that the button does not look the best. I'll leave styling up to someone else.

api/src/config/deploy.js
aksh0001: Yep, `scholly` resolves to the IP of the container.

aksh0001: Read https://docs.docker.com/compose/networking/

```
y default Compose sets up a single network for your app. Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by them at a **hostname identical to the container name.**
```

aksh0001:
> Thanks for fixing/cleaning/adding these things! Just got a few suggestions and questions (out of my own curiosity)...
> 
> Also, if it's feasible it'd be great if we can split the PR in the future (if there's a lot of changes being made). Looking at this figure just makes me want to procrastinate sometimes , although most of these files are just small changes or just package.json files
> Screen Shot 2021-08-17 at 9 50 47 pm

You're right sorry about the size. I couldn't find a decent split for this PR as all changes were critical that they happen together. But I guess I should have split up the refactors into a separate PR. Will do next time

api/src/controllers/team.js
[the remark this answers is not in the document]
vgoh1: Do you mean you intended scholly to be fetching the data, and not the Researchify backend which passes it onto scholly?

client/src/components/dashboard/deploy/DeployPage.js
vgoh1: I have a ticket on winter's board to refactor the deploy page anyway, so I'll put that in there for someone to handle it.

vgoh1 (approved):

rpra0006 (approved):

Dockerrun.aws.json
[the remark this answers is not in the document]
kdav108: Yep no stress, just wanted to confirm

api/src/controllers/team.js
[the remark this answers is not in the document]
kdav108: I think he means the Reserachify backend should not be requiring the frontend to pass this information:
```
const { ghToken, teamPublications, teamInfo, teamMembers } = req.body;
```
Which I agree, the backend can fetch that information itself. Currently the frontend queries backend for these data and then passes it backend again, so no point of having that circle. 

Maybe @aksh0001, if I'm understanding correctly, then can we place these 2 lines of comments above line 241. That would make it easier to understand.

scholly/Dockerfile
kdav108: Right great explanation. So if I understand correctly we could also alternatively use something like `chmod` or `chown` to give `node` permission to `/home/node/app` right?

aksh0001: Yep, that’d work too.

api/src/controllers/team.js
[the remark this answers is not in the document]
aksh0001: Yep that’s what I’m trying to convey. It doesn’t make much sense for the client to fetch this data from the backend to then send it back.
I think of all the refactors this should be a priority.

@kdav108 will do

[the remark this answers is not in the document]
aksh0001: Done.

kdav108 (approved):
LGTM!

api/src/config/deploy.js
```diff
@@ -1,14 +1,28 @@
...
+  githubClientId = 'bf0b026536ac6923e06e';
+} else {
+  schollyHost = 'http://localhost:8000';
+  githubClientId = '0a58958e33b392f837b6';
+}
+
+// Client secret of the configured GitHub OAuth application.
 const githubClientSecret = process.env.GH_CLIENT_SECRET;
```
ssha0054, line 22: Was just wondering how we can get this credential.

ssha0054 (approved):
Looks good!

api/src/config/deploy.js
aksh0001: Thanks for bringing this up. Added to the drive.

PR #167, opened 2021-08-17
Enable stricter ESLint checks for our API and Scholly via AirBnB style guide

aksh0001:
# Description 

Extend our current ESLint configuration to enforce more stricter checks to ensure code quality using AirBnB's style guide.

- See the AirBnB style guide: https://github.com/airbnb/javascript
- We could also choose to extend the Google JS style guide, but I prefer the AirBnB because it has more checks. See the comparison here: https://betterprogramming.pub/comparing-the-top-three-style-guides-and-setting-them-up-with-eslint-98ea0d2fc5b7

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
Our ESLint checks were not rigorous enough on the API and Scholly service that led to reviewers leaving comments that could have been addressed by our Static Analyzer.

# Solution description
Use the AirBnB ESLint configuration to enforce stricter style checks.

# Tests 
All unit tests pass? T

Builds successfully? T

# Risk
Medium. There's plenty of refactors in this change, when reviewing please check for any regressions that might have come up.


aksh0001:
FYI, I'll be requesting reviews after the Scholly AWS deployment PR is merged since these checks also run on those changes: https://github.com/Researchify/Researchify/pull/163

jfan0010 (approved):
Dumb question: why are we removing default users in `seederScript.js` while keeping others?

aksh0001:
> Dumb question: why are we removing default users in `seederScript.js` while keeping others?

I believe the `User` model is no longer used in favor of the broader `Team` model. I had to remove it because ESLint was throwing an error due to an unsolved reference to `require('./models/user.model')`

jfan0010:
> > Dumb question: why are we removing default users in `seederScript.js` while keeping others?
> 
> I believe the `User` model is no longer used in favor of the broader `Team` model. I had to remove it because ESLint was throwing an error due to an unsolved reference to `require('./models/user.model')`

My bad, I thought it was the team model

api/src/controllers/publications.js
```diff
@@ -185,13 +180,11 @@ async function getGoogleScholarPublications(req, res) {
 
     await browser.close();
 
-    await Promise.all(links.map((x) => scrapeGoogleScholar(x))).then((pub) =>
-      publications.push(...pub)
-    );
+    await Promise.all(links.map((x) => scrapeGoogleScholar(x))).then((pub) => publications.push(...pub));
```
vgoh1, line 183: Not an issue, but I'm surprised you're able to modify arrays when they're initialized to const (L155)? I always thought you couldn't.

api/src/controllers/publications.js
```diff
@@ -143,21 +139,20 @@ function readAllPublicationsByTeam(req, res, next) {
...
-    playwrightConfig.sortBySuffix;
+  const { noOfDummyLinks } = playwrightConfig;
+  const { pageSize } = playwrightConfig;
+  const url = playwrightConfig.baseUrl
+    + author
+    + playwrightConfig.startSuffix
+    + startFrom
+    + playwrightConfig.pageSizeSuffix
+    + pageSize
+    + playwrightConfig.sortBySuffix;
```
vgoh1, line 153: Another thing I noticed is that the AirBnB style guide conflicts with prettier sometimes and this block is an example.  Can we somehow configure prettier to follow this style if we're going to be enforcing it?

api/src/controllers/theme.js
```diff
@@ -5,10 +5,10 @@
...
 
 /**
- * 
- * @param {*} req 
- * @param {*} res 
- * @returns 
+ *
+ * @param {*} req
+ * @param {*} res
+ * @returns
```
vgoh1, line 11: This probably needs to be filled in later.

vgoh1 (approved):

api/src/controllers/publications.js
aksh0001: yeah I believe you cannot re-assign them, hence `const`, but you are able to mutate the contents.

api/src/controllers/publications.js
aksh0001: Not too sure, but for VSCode users, you can enable this extension https://daveceddia.com/vscode-use-eslintrc/#:~:text=Configure%20VSCode%20Settings%20to%20use%20ESLint%20for%20Formatting&text=Click%20that%20tiny%20icon%20in,paper%20with%20a%20little%20arrow.&text=The%20first%20one%20turns%20on,it%2C%20we're%20done.

Similar exists for Webstorm users. And re: prettier, we may need someone to look into that,

api/src/controllers/theme.js
aksh0001: Added a story to winter's board as a stopgap since I believe one of you worked on theme:
https://trello.com/b/5qazh84D/team-winter

PR #176, opened 2021-08-22
Import publications refactor

vgoh1:
# Description 
Initially this was in another branch for figuring out why import wasn't working in prod but I was clogging up a lot of commits in that branch and ended up breaking the staging env......so here are just the code changes for the fix.

- Refactor validate gscholarid middleware to use try/catch
- Add teamid validation for importing publications
- Refactor the import fail page to have a proper message since the error state gets cleared and shows up as a toast

# Type of change 
- [x] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
See description

# Solution description
Should be self explanatory

# Tests 
Please run the import functionality just in case I overlooked something.

Import fail looks like this now:
image

N/A

# Risk
Low, just the import functionality

### DIFF
--- a/api/src/middleware/publications.js
+++ b/api/src/middleware/publications.js
@@ -97,7 +97,7 @@ const createPublicationValidation = [
 async function validateAuthorId(req, res, next) {
   const { gScholarUserId: _id } = req.params;
   if (_id.length !== 12) {
-    next(
+    return next(
       fillErrorObject(400, 'Validation error', [
         'Google Scholar User ID needs to be 12 characters long',
       ]),
@@ -105,11 +105,11 @@ async function validateAuthorId(req, res, next) {
   }
 
   try {
-    await axios(`${playwrightConfig.baseUrl}${_id}`);
-    next();
+    await axios.get(`${playwrightConfig.baseUrl}${_id}`);
+    return next();
   } catch (error) {
     // if you mess around with the user id you only get 404
-    next(
+    return next(
       fillErrorObject(error.response.status, 'Validation error', [
         'No Google Scholar user profile found with the given id',
       ]),
--- a/api/src/middleware/publications.js
+++ b/api/src/middleware/publications.js
@@ -4,7 +4,7 @@
 const { body, validationResult } = require('express-validator');
 const axios = require('axios');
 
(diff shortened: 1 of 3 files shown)


jfan0010 (approved):

api/src/middleware/publications.js
```diff
@@ -103,19 +103,18 @@ async function validateAuthorId(req, res, next) {
...
-        fillErrorObject(
-          error.response.status,
-          'Validation error',
-          ['No Google Scholar user profile found with the given id'],
-        ),
-      );
-    });
+
+  try {
+    await axios(`${playwrightConfig.baseUrl}${_id}`);
```
aksh0001, line 108: Making this read `axios.get()` should be more readable.
So the error was because we weren't using `playwrightConfig.baseUrl`?

api/src/middleware/publications.js
```diff
@@ -103,19 +103,18 @@ async function validateAuthorId(req, res, next) {
...
+    next();
+  } catch (error) {
+    // if you mess around with the user id you only get 404
+    next(
```
aksh0001, line 112: prefix with a `return` to return early from this handler and avoid executing any unintentional code if added below.

aksh0001 (approved):

rpra0006 (approved):

api/src/middleware/publications.js
vgoh1: Will do. I think it could've been part of it since it was using .com.sg but it still doesn't make sense why it would start returning 403 now and not before. Hoping this will fix it in prod.

api/src/routes/publications.js
```diff
@@ -32,8 +32,9 @@ publicationsRouter.post(
 
 publicationsRouter.get(
   '/import/:gScholarUserId/:startFrom/validate/:team_id',
-  authMiddleware.cookieJwtAuth,
+  // authMiddleware.cookieJwtAuth,
```
aksh0001, line 35: please remove if not needed

vgoh1: oops, i always forget this. will fix on main

PR #178, opened 2021-08-23
Added toast error when user failed to log in GitHub

jfan0010:
# Description 
as title

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
previously console.error

# Solution description
as above

image

### Provide steps in detail for reviewers to reproduce the expected result of the code changes.
when logging in, close the github window

# Risk
Tiny


client/src/components/dashboard/deploy/DeployPage.js
```diff
@@ -11,7 +12,6 @@ import './DeployPage.css';
 const DeployPage = ({ teamId }) => {
   const dispatch = useDispatch();
   const loading = useSelector((state) => state.deploy.loading);
-  // TODO: refactor this into useState
```
aksh0001, line 14: So we're not going with `useState`?

aksh0001 (approved):

rsal0007 (approved):

client/src/components/dashboard/deploy/DeployPage.js
jfan0010: cause before we wanna check if the user has added any page before deploy, but now we made `client homepage` mandatory, so I think we don't need `useState` atm

ssha0054 (approved):

PR #186, opened 2021-08-25
Configure eslint to follow airbnb style guide for client

vgoh1:
# Description 
See title. Scholly/base will come in another PR because this is already huge.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
Client code isn't very consistent so we should enforce a style guide.

# Solution description
I've suppressed a lot of rules because fixing all of those in this one PR would take a long time, but I took note which ones were important enough to refactor soon-ish and which ones were low-prio. 

# Tests 
If people could check that the functionality still works and call out anything that broke in the process that would be appreciated!

N/A

# Risk
High


client/.eslintrc.json
```diff
@@ -0,0 +1,37 @@
...
+        "node": true
+    },
+    "extends": ["eslint:recommended", "airbnb", "prettier"],
+    "parserOptions": {
+        "ecmaVersion": 12
+    },
+    "rules": {
+        "max-len": "off",
+        "no-underscore-dangle": "off",
+        "camelcase": "off",
```
aksh0001, line 14: not a question, just an observation: I'm surprised that we're not actually following camelcase

vgoh1: Actually, I just copied this file from /api so the first few rules got carried over here. With the rule on, there's only 1 case of it - several follow up PRs are probably going to come next PI to go through these rules and fix violations.

aksh0001 (approved):

rpra0006 (approved):
Code still runs for me

client/src/components/publications/Publications.js
```diff
@@ -83,7 +84,12 @@ const Publications = () => {
         {loading ? (
           <Spinner animation="border" />
         ) : (
-          <h4>Total of {teamPublications.length} publications</h4>
+          <h4>
+            Total of
+            {teamPublications.length}
+            {' '}
```
jfan0010, line 90: need a space between `line 88` and `line 89`

jfan0010 (approved):
LGTM

client/src/components/publications/Publications.js
vgoh1: thanks for the catch

PR #188, opened 2021-08-25
Added 2 Numbers to Website model to store selected colour & layout as…

AkashSaggar:
Added 2 Numbers to Website model to store selected colour & layout as theme info

# Description 
The Theme and Template models are not needed as all we need for theme info are 2 integers.
One to specify the selected colour option (from 1 to 3, or maybe more as we add more colours).
One to specify the selected layout option (from 1 to 3, or maybe more as we add more layouts).
The themeId field in the Team model is also not needed.

However these 3 unneeded things (2x models and 1x field) will be deleted later.
For now I am just adding the fields we actually want to use to the website model.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem

# Solution description

# Tests 
All unit tests pass?

Builds successfully?

# Risk


api/src/models/website.model.js
```diff
@@ -37,6 +37,16 @@ const websiteSchema = new mongoose.Schema(
...
+    colour: {
+      type: Number,
+      required: true,
+      default: 1,
+    },
+    layout: {
+      type: Number,
+      required: true,
+      default: 1,
+    },
```
aksh0001, line 49: If these are marked as `required` then it might be a good idea to implement the controller functionality in this PR as well (which should be simply updating these two fields based). Also since these are essentially enum constants we should follow the convention we've been using for enum fields in other models.

@pmyip3 can be a helpful source of information here.

https://github.com/Researchify/Researchify/blob/main/api/src/models/website.model.js#L32

aksh0001:
Also, curious: if we use the existing colors set in the `Theme` model:https://github.com/Researchify/Researchify/blob/main/api/src/models/editor/theme.model.js#L12-L30

Don't we have more flexibility in choosing any color we want? What advantage is there if we're having a limited choice?

PR #190, opened 2021-08-26
Make images for theme colour and layout clickable

ssha0054:
# Description 
On the Theme selector page, the images are currently not clickable. In order to select a theme colour or layout, you need to select the checkbox. This PR makes these images clickable so you now have another way of selecting a theme and layout.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
Images in the theme selector were not clickable.

# Solution description
Made the images for both theme and layout have onClick actions which result in the corresponding radio button to be checked whilst also ensuring the correct information regarding theme and layout is stored in the state.

# Tests 
All unit tests pass? Yes

Builds successfully? Yes

Before:
image

After clicking on the image for layout 3 and the image for the black and white theme:
image

# Risk
Low

### DIFF
--- a/client/src/actions/auth.js
+++ b/client/src/actions/auth.js
@@ -8,7 +8,6 @@ import {
   LOG_OUT,
   LOG_IN_FAIL,
   FETCH_TEAM_INFO,
-  CLEAR_NOTIFICATION,
   FETCH_WEBSITE_INFO,
 } from './types';
 import { errorActionGlobalCreator } from '../notification/notificationReduxFunctions';
@@ -58,10 +57,6 @@ export const logout = () => async (dispatch) => {
     await api.logoutTeam();
     dispatch({
       type: LOG_OUT,
-    });
-
-    dispatch({
-      type: CLEAR_NOTIFICATION,
     });
   } catch (err) {
     dispatch(errorActionGlobalCreator(err));
--- a/client/src/reducers/index.js
+++ b/client/src/reducers/index.js
@@ -1,5 +1,6 @@
 /**
- * This index file will export combined reducers.
+ * This index file will export combined reducers
+ * When the logout action is dispatched, all reducers will be initialzed to thir initial state
  */
 import { combineReducers } from 'redux';
 
@@ -13,8 +14,9 @@ import notificationReducer from './notificationReducer';
 import deployReducer from './deployReducer';
 import achievementsReducer from './achievementsReducer';
 import homepageReducer from './homepageReducer';
+import { LOG_OUT } from '../actions/types';
 
-export default combineReducers({
+const appReducer =  combineReducers({
   notification: notificationReducer,
   auth: authReducer,
   publications: publicationsReducer,
@@ -26,3 +28,12 @@ export default combineReducers({
   achievements: achievementsReducer, 
   homepage: homepageReducer,
 });
+
+const rootReducer = (state, action) => {
+  if (action.type === LOG_OUT) {
+    return appReducer(undefined, action)
+  }
+  return appReducer(state, action)
+}
+
+export default rootReducer


client/src/components/dashboard/TemplateSelector.js
```diff
@@ -88,30 +87,33 @@ const TemplateSelector = (props) => {
...
+              className="theme-icon theme-1-icon" 
+              onClick={() => updateSelections('theme', 1)} 
+            />
             <Form.Check
               inline
               type="radio"
-              name="theme"
-              value="theme2"
-              onChange={updateForm}
+              onChange={() => updateSelections('theme', 2)}
```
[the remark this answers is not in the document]
pmyip3: Not quite sure why name and values do not work but figure out that the id props work.
image

client/src/components/dashboard/TemplateSelector.js
```diff
@@ -88,30 +87,33 @@ const TemplateSelector = (props) => {
...
-              value="theme3"
-              onChange={updateForm}
+              onChange={() => updateSelections('theme', 3)}
               checked={theme === 3}
             />
-            <div className="theme-icon theme-3-icon" />
+            <div 
+              className="theme-icon theme-3-icon" 
+              onClick={() => updateSelections('theme', 3)} 
+              />
```
pmyip3, line 116: Just a suggestion for improvement is it possible to combine the `<Form.Check> `and `<div>` into a single div so we only handle the onClick function in a single place instead of 2?

pmyip3 (commented):
Another suggestion: might have it in a separate ticket though as I am not quite sure if the numbers of color/ layout would be increases or not. If that is the case, this files might become large, so would be nice if we can have another file to store the relevant layout and color data like what we have in SidebarData.js

client/src/components/dashboard/TemplateSelector.js
jfan0010: @ssha0054  Good news, I found a way to wrap them, see [here](https://stackoverflow.com/questions/19412204/how-do-you-make-the-radio-button-text-to-be-clickable-too/54573527)

client/src/components/dashboard/TemplateSelector.js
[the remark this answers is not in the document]
jfan0010: @ssha0054 just found out that using `label` you can wrap them together, and doesn't need to implement two `onClick` or `onChange` in one component. See: https://stackoverflow.com/questions/19412204/how-do-you-make-the-radio-button-text-to-be-clickable-too/54573527

ssha0054 (commented):
**Changes made:** Used a `label` to wrap the radio buttons and images so now only one `onChange`/`onClick` call is needed. Also removed hardcoding in the values when calling the `updateSelections` function, however had some issue with the `value` being used in the case statements so decided to use the `id` instead. These changes have also made the theme image to now be below the radio button instead of being on the side. Refer to pic below: 
image
Not sure how to fix this atm but will make sure it is back to how it was originally before merging.

client/src/components/dashboard/TemplateSelector.js
```diff
@@ -26,12 +26,14 @@ const TemplateSelector = (props) => {
   });
   const [theme, setTheme] = useState(1);
 
-  const updateForm = (form) => {
-    const { name, value } = form.target;
+  const updateSelections = (form) => {
+
```
pmyip3, line 30: empty lines in line 30 and 32

pmyip3 (approved):

aksh0001 (approved):
Will defer to @jfan0010 for the React practices.

aksh0001:
Actually, I wanted to add that this entire UI may need to be refactored as we're thinking of implementing a toggle between light/dark theme as opposed to color selection.
@vgoh1 @sanathmahesh @rsal0007 @melvinalex7 @melvinalex77

ssha0054:
> Actually, I wanted to add that this entire UI may need to be refactored as we're thinking of implementing a toggle between light/dark theme as opposed to color selection.
> @vgoh1 @sanathmahesh @rsal0007 @melvinalex7 @melvinalex77

In that case, should this PR still get merged into main as the theme layout images being clickable would still be useful for later (even with any changes to this UI)? Or would it be better to just close this PR and let whoever works on the implementation of the toggle to also implement the ability to click on any images, if required.

vgoh1:
@ssha0054 I think it's worth merging because the redesign is still far away in the future. The theme layout pictures should still be useful.

vgoh1 (approved):
Helped fix a bug (L52 in TemplateSelector.js) that caused the site to crash when you hit the create website button in the theme selector. :) LGTM otherwise.

PR #194, opened 2021-08-29
Deploy page fail when no webpage found in db

pmyip3:
# Description 
A bug found: failing to deploy to GitHub when no webpage info of a team is found in the db 
Read more: https://trello.com/c/tbBk8hXN 

# Type of change 
- [x] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
In the current code, when no webpage info is found for a specific team, it returns a 404 code with a string. Thus, for a team with no websiteInfo created in db, an error would occur when trying to deploy to github 
image

# Solution description
A mongoose post middleware is created which automatically creates a website and homepage document after a team has registered (Thank to @aksh0001 )
See more post middleware: https://stackoverflow.com/a/57971800/15507541    https://mongoosejs.com/docs/middleware.html#post
**Note: For the newly created team, the error should be fixed. However, for the existing team, the error would still exist. Pls register a new team to prevent the error mentioned above.** 

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

- Registrate a new team and double-check if the webpage and homepage documents of that team are created, then try to deploy the page to github. 
- Pls let me know if any bugs exist in adding/ deleting web pages functionality as I've also refactored these 2 endpoints. 

https://trello.com/c/tbBk8hXN 

# Risk
Low to Medium. Modified a few endpoints, might affect other functions which use this endpoint I am not aware of.


cyr17 (approved):
Legit!

api/src/controllers/website.js
```diff
@@ -10,25 +10,21 @@ const { fillErrorObject } = require('../middleware/error');
...
-    })
-    .catch((error) => res.status(500).json(`Server Error: ${error.message}`));
+async function getWebPageDetails(req, res, next) {
+  const { team_id } = req.params;
+  try {
+    const foundWebsiteInfo = await Website.findOne({ teamId: team_id });
+    if (foundWebsiteInfo) {
+      return res.status(200).json(foundWebsiteInfo);
+    }
+    const emptyWebsiteInfo = { teamId: team_id, pages: [] };
```
aksh0001, line 23: I think a more scalable solution is to have an invariant where an empty list of pages is associated with a `Team` by default, so we don't have to interfere and force returning an empty page list to fix this issue. What do you reckon?
What about when a user first signs up, we initialize their webpages to empty for them? Would this be a cleaner approach?

aksh0001: This basically ensures all state associated with a user's page is default initialized to appropriate values

aksh0001: To make this even more scalable, we could consider implementing db triggers: https://www.google.com/search?q=db+trigger+mongodb&oq=db+trigger+mongodb&aqs=chrome..69i57j0i22i30j0i10i22i30j0i22i30l3.3014j0j7&client=ubuntu&sourceid=chrome&ie=UTF-8

so that the appropriate models get auto updated when a user signs up

vgoh1 (approved):
Deployed site and page selector works fine on my end. The DB trigger is neat!

api/src/controllers/homepage.js
```diff
@@ -18,9 +19,7 @@ async function getHomepage(req, res, next) {
     if (foundHomepage) {
       return res.status(200).json(foundHomepage);
     }
-    // note: each string in `aboutUs` is a paragraph
-    const emptyHomepage = { teamId: team_id, aboutUs: [''] };
-    return res.status(200).json(emptyHomepage);
+    return next(fillErrorObject(404, 'Validation error', ['No homepage found with the given team_id']));
```
aksh0001, line 22: Thanks for writing those triggers! I can imagine one downside where, if the trigger is somehow removed or if we migrate dbs, we can run into problems when we run our code. I'd add documentation somewhere to note that we have a trigger to updated collection x based on insertions to the `Team` model. Maybe the best place is to add a comment in the schema model files of affected collections, like `Homepage`, `Website`, etc.

api/src/controllers/website.js
```diff
@@ -10,90 +10,44 @@ const { fillErrorObject } = require('../middleware/error');
...
+  const { team_id } = req.params;
+  try {
+    const foundWebsiteInfo = await Website.findOne({ teamId: team_id });
+    if (foundWebsiteInfo) {
+      return res.status(200).json(foundWebsiteInfo);
+    }
+    return next(fillErrorObject(404, 'Validation error', ['No webpage detail found with the given team_id']));
+  } catch (err) {
+    return next(fillErrorObject(500, 'Server error', [err.errors]));
+  }
```
aksh0001, line 27: Ditto, could you document that that this model is associated with a trigger that will automatically create an empty webpage if a team signs up.

aksh0001: In the model files would be the best place I reckon, similar to the above.

aksh0001:
Do you know if we can programmatically create that trigger you just made @mable-yip @pmyip3 ? It may be the solution to the problems I brought up.

pmyip3:
> Do you know if we can programmatically create that trigger you just made @mable-yip @pmyip3 ? It may be the solution to the problems I brought up.
Don't have any idea. But I did some research on that, instead of using functions to create the trigger, MongoDB also allows using Amazon EventBridge. I don't know much about AWS tho, but seems like we can configure the trigger in AWS and then provide the AWS account ID to MongoDB. 
https://docs.atlas.mongodb.com/triggers/eventbridge/

aksh0001:
@pmyip3 @mable-yip I think I may have found a neater solution that can avoid triggers + clean up the original approach. We can apparently implement mongoose middleware to act like triggers on the `Model` level. I've pushed up a change to this branch https://github.com/Researchify/Researchify/commit/ff4ca6efc503e9e679905076b8ca76c73debceed so you can have a look.

IMO, it's pretty neat. Could you test it out and lmk what you think?

References:
https://stackoverflow.com/questions/57971183/is-there-a-way-to-create-after-update-triggers-or-events-in-mongoose
https://mongoosejs.com/docs/middleware.html#post
https://mongoosejs.com/docs/api.html#schema_Schema-post

aksh0001:
> @pmyip3 @mable-yip I think I may have found a neater solution that can avoid triggers + clean up the original approach. We can apparently implement mongoose middleware to act like triggers on the `Model` level. I've pushed up a change to this branch [ff4ca6e](https://github.com/Researchify/Researchify/commit/ff4ca6efc503e9e679905076b8ca76c73debceed) so you can have a look.
> 
> IMO, it's pretty neat. Could you test it out and lmk what you think?
> 
> References:
> https://stackoverflow.com/questions/57971183/is-there-a-way-to-create-after-update-triggers-or-events-in-mongoose
> https://mongoosejs.com/docs/middleware.html#post
> https://mongoosejs.com/docs/api.html#schema_Schema-post

@vgoh1 @rpra0006 tagging so you can pitch in on this approach, if you want.

api/src/models/team.model.js
```diff
@@ -50,6 +67,18 @@ const teamSchema = new mongoose.Schema(
   { timestamps: true },
 );
 
+// Post middleware/trigger to create associated documents when a team has
+// registered. See: https://stackoverflow.com/a/57971800/15507541.
+// https://mongoosejs.com/docs/middleware.html#post
+teamSchema.post('save', async (doc) => {
+  try {
+    await Website.create({ teamId: doc._id });
+    await Homepage.create({ teamId: doc._id });
```
aksh0001, line 76: I believe with these additions, these `model` changes are the only ones that we need to make. LMK if it works for you, and feel free to modify this code as I haven't tested thoroughly - it was more of ad hoc research.

aksh0001: Could you also cross-check the documents in the db? I just did some minor verification and it appears we have two associated docs creates for each `Website` and `Homepage` models when I tried registering. Pretty weird behavior, but I may have done something wrong.

aksh0001: Ahhh okay one of those created may have been from the trigger we made earlier. But we should be able to do without it if we go with this approach.

pmyip3: Thanks for that! It works for me. I'll test it properly and go through other controllers as well.

pmyip3: I have removed the trigger in MongoDB and other controllers look good to me. Free feel to review this PR again.

aksh0001 (approved):
LGTM, approval.
The code looks a lot cleaner! Thanks for this.

vgoh1 (approved):
The middleware relating to the docs is super cool, didn't know that was possible. Love how neat the solution is now.

PR #195, opened 2021-08-30
Fix import publications bug with category types

vgoh1:
# Description 
The refactoring of category types for the publications from PR#171 broke some of the code from import publications logic in the backend which means every imported publication gets classified as 'Other'. 

# Type of change 
- [x] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
See description
image

# Solution description
Backend logic had to be fixed to not compare using uppercase and use comparisons for an array since `categoryType` is not an object anymore
The frontend/scholly base code also needed to be changed to not use uppercase in comparisons, because this broke the publications layout by category. This also broke add publications since it's using the `categoryTypes` object.
Also did some refactoring so that a space would appear after the colon in the publications details

# Tests 
Try importing and checking that the publications get classified properly, and change the layout to sort by category. Also try deploying the site and checking that sort by category works.

N/A

# Risk
Medium, scholly base, client publications page, import logic in backend

### DIFF
--- a/api/src/controllers/publications.js
+++ b/api/src/controllers/publications.js
@@ -234,9 +234,9 @@ async function scrapeGoogleScholar(url) {
 
   // TODO: this logic depends on the order of the fields,
   // which will differ based on the info of the publication, can be improved
-  let type = fields[2].toUpperCase();
+  let type = fields[2];
   let categoryTitle;
-  if (!(type in categoryTypes)) {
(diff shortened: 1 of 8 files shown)


pmyip3 (approved):

jfan0010 (approved):
LGTM

rpra0006 (approved):

PR #196, opened 2021-08-31
Eslint config for scholly base

vgoh1:
# Description 
See title. Luckily not a huge PR this time and not that many rules to suppress. I also made added prettier to the backend and scholly base eslint config so that the airbnb config won't complain about it.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
See title.

# Solution description
Same as PR#186

# Tests 
Please help regression test by deploying to GH pages and see if the site loads, as well as running the scholly base react app locally. 

N/A

# Risk
Medium, scholly base can fail to build if eslint throws anything


pmyip3 (approved):
Both deploying and scholly base works for me

api/.eslintrc.json
```diff
@@ -4,7 +4,7 @@
         "es2021": true,
         "node": true
     },
-    "extends": ["eslint:recommended", "airbnb-base"],
+    "extends": ["eslint:recommended", "airbnb-base", "prettier"],
```
aksh0001, line 7: Curiou: what does `prettier` have to do with eslint config?

scholly/base/src/components/publications/PublicationPage.js
```diff
@@ -10,49 +10,45 @@ import Container from 'react-bootstrap/Container';
...
+  //     <Fragment>
+  //       <Container className="pages-top-padding text-center mt-3 mb-3">
+  //         <div className="publication-pg-title">Our Publications</div>
+  //       </Container>
+  //       <Container fluid>
+  //         <Publications />
+  //       </Container>
+  //     </Fragment>
+  //   )}
+  // </Fragment>
```
aksh0001, line 43: Can this commented section be removed?

scholly/base/src/global/data.js
```diff
@@ -46,7 +46,6 @@ if (!env.REACT_APP_DEBUG) {
     ? JSON.parse(env.REACT_APP_WEB_PAGES)
     : [];
 } else {
-  console.log('Running in DEBUG mode, hence using fake Team data');
```
aksh0001, line 49: Actually, would prefer to leave the `console.log` here as it's useful to know that we're running in Debug mode, or else it can be confusing whether we're using real or fake data. Maybe  we could also modify the log to read `..., hence using fake data`, as we're not only using fake Team data.

aksh0001 (approved):
LGTM, thanks for adding this!

scholly/base/src/global/data.js
vgoh1: Sure, added this line back.

scholly/base/src/components/publications/PublicationPage.js
vgoh1: This is still here because Jie plans on adding the twitter box to the deployed website, at the time he couldn't figure out a way to make it look nice so it's been commented out. iirc there's plans to use this code later so I'm leaving this as is.

api/.eslintrc.json
vgoh1: So if prettier is added at the end here, when you use prettier for formatting and it conflicts with airbnb's config, npm run lint won't throw an error about it. You can test it with this part of the code: https://github.com/Researchify/Researchify/pull/167#discussion_r692588868
If you run prettier on the file and run `npm run lint` without prettier in the extends array, it'll complain about formatting. 
@kdav108 figured this out. :)

scholly/base/src/components/publications/PublicationPage.js
jfan0010: I will change it asap (probably this weekend) if we're still using this layout without much amendment.

jfan0010 (approved):
Works on mine

PR #198, opened 2021-08-31
Team Data Reset Integration

cyr17:
# Description 
A summary of the change 
Implemented Delete Account functionality, which provides the user with two options 
1: Delete account data
2: Reset account data wtihout delete ghpages
The functionality by default deletes the GH repository but wouldnt delete it if not logged onto github in the currently active session.

Enhancement:
This PR has been modified such that it doesnt rely on calling multiple dispatch events which were responsible for resetting the redux states, this was dropped as it would be harder to keep track and modify in the future. This is now replaced by a upper level dispatch call equivalent to a logout action type which resets everything instead.

TODO:
As this PR was out of scope from what it initially was, we have added a TODO list for further enhancements.
Access tokens not cleared in local storage - this is out of the scope and we shouldn't store the access tokens in the local storage as it can be easily misused.

# Type of change 
- [x] Bug fix
- [x] New feature
- [x] Refactoring
- [x] Documentation 

# Problem
An option to delete the already deployed GH pages of a user and to delete the collected user data from the system.

# Solution description

# Tests 
All unit tests pass?
yes
Builds successfully?
yes

Attach a screenshot of the expected result (e.g. UI of the website/ postman API call result/ database changes)

image
image
image

Deleting repo without having github pages 
image

Deleting repo without logging in 
image

Deploy a website using ur Github account
Click on the Team Profile button --> Delete Account --> relevant option

https://trello.com/c/Fl6kQMnX

# Risk
medium
The access token is not cleared from the local storage on every log out. This could become a high-potential risk as it could help tamper with the previous users GitHub repos if there are multiple users logging in via the same device.

### DIFF
--- a/client/src/components/achievements/achievementsOnboarding.js
+++ b/client/src/components/achievements/achievementsOnboarding.js
@@ -15,17 +15,17 @@ const AchievementsPageWalkthrough = () => {
       {
         element: '#add-achievement-button',
         title: 'Add Achievement',
-        intro: 'Click here to add a new achievement manually',
+        intro: 'Click here to add a new achievement.',
       },
       {
         element: '#delete-icon',
         title: 'Deleting selected Achievements',
-        intro: 'Click here to delete all selected achievements',
+        intro: 'Click here to delete all selected achievements.',
       },
       {
         element: '#select-achievements-checkbox',
         title: 'Selecting Achievements',
-        intro: 'Click on this checkbox to select all achievements',
+        intro: 'Click on this checkbox to select all achievements.',
       },
     ],
   });
@@ -58,7 +58,7 @@ const AchievementsEditDeleteWalkthrough = () => {
     steps: [
       {
         element: '#achievement-checkbox',
-        intro: 'Click on this checkbox to select the added achievement.',
+        intro: 'Click on this checkbox to select the achievement.',
         position: 'bottom',
       },
       {
@@ -98,12 +98,12 @@ const AchievementsFormWalkthrough = () => {
       {
         element: '#form',
         title: 'New Achievement',
-        intro: 'Please ensure that: all fields are filled',
+        intro: 'Please ensure that all fields are filled.',
       },
       {
         element: '#achieve-title',
         title: 'Achievement Title',
-        intro: 'Please ensure that the title is 3-60 characters',
+        intro: 'Please ensure that the title is 3-60 characters.',
       },
       {
         element: '#year',
@@ -113,7 +113,7 @@ const AchievementsFormWalkthrough = () => {
       {
         element: '#desc',
         title: 'Achievement Description',
-        intro: 'Please ensure that the description is 5-500 characters',
+        intro: 'Please ensure that the description is 5-500 characters.',
       },
     ],
   });
--- a/client/src/components/publications/PublicationPage.js
+++ b/client/src/components/publications/PublicationPage.js
@@ -10,6 +10,7 @@ import PublicationsEditor from './publicationsLayout/PublicationsEditor';
 import PublicationForm from './form/PublicationForm';
 import ImportForm from './form/ImportForm';
 import { sortingOptions } from '../../config/publications';
+import PublicationsPageWalkthrough from './publicationsOnboarding';
 
 const sortPublications = (publicationToBeSorted, option) => {
   switch (option) {
@@ -80,7 +81,10 @@ const PublicationPage = () => {
 
   return (
     <div className="publicationPageContainer">
-      <h2>Publications</h2>
+      <div style={{ display: 'flex' }}>
+        <h2 style={{ marginRight: '10px' }}> Publications </h2>
+        <PublicationsPageWalkthrough />
+      </div>
       <PublicationsEditor
         options={options}
         setOptions={setOptions}
--- a/client/src/components/publications/form/CategoryPage.js
+++ b/client/src/components/publications/form/CategoryPage.js
@@ -17,6 +17,7 @@ import { PropTypes } from 'prop-types';
 import { categoryTypes } from '../../../config/publications';
 import '../publications.css';
 import { PrimaryButton, DangerButton } from '../../shared/styledComponents';
+import { PublicationsManualCategoryFormWalkthrough } from '../publicationsOnboarding';
 
 const CategoryPage = ({
   next, prev, data, closeModal,
@@ -73,7 +74,8 @@ const CategoryPage = ({
           setValues,
         }) => (
           <Form noValidate onSubmit={handleSubmit}>
-            <Form.Group>
+            <PublicationsManualCategoryFormWalkthrough />
+            <Form.Group id="publication-category">
               <div className="text-center">
                 <ButtonGroup toggle>
                   {Object.keys(categoryTypes).map((category) => (
@@ -98,7 +100,7 @@ const CategoryPage = ({
               </div>
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-category-title">
               <Form.Label>
                 {' '}
                 {Object.keys(categoryTypes)
@@ -130,7 +132,7 @@ const CategoryPage = ({
               </Form.Control.Feedback>
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-volume">
               <Form.Label> Volume </Form.Label>
               <Form.Control
                 className="placeholder-text"
@@ -142,7 +144,7 @@ const CategoryPage = ({
               />
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-issue">
               <Form.Label> Issue </Form.Label>
               <Form.Control
                 className="placeholder-text"
@@ -154,7 +156,7 @@ const CategoryPage = ({
               />
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-pages">
               <Form.Label> Pages </Form.Label>
               <Form.Control
                 className="placeholder-text"
@@ -166,7 +168,7 @@ const CategoryPage = ({
               />
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-publisher">
               <Form.Label> Publisher </Form.Label>
               <Form.Control
                 className="placeholder-text"
--- a/client/src/components/publications/form/MainInfoPage.js
+++ b/client/src/components/publications/form/MainInfoPage.js
@@ -16,6 +16,7 @@ import {
 import '../publications.css';
 import { PropTypes } from 'prop-types';
 import { PrimaryButton, DangerButton } from '../../shared/styledComponents';
+import { PublicationsManualFormWalkthrough } from '../publicationsOnboarding';
 
 const MainInfoPage = ({
   next, data, type, pub, closeModal,
@@ -106,8 +107,9 @@ const MainInfoPage = ({
           errors,
           setValues,
         }) => (
-          <Form noValidate onSubmit={handleSubmit}>
-            <Form.Group>
+          <Form noValidate onSubmit={handleSubmit} id="publication-title-form">
+            <PublicationsManualFormWalkthrough />
+            <Form.Group id="publication-title">
               <Form.Label>
                 Publication Title
                 {' '}
@@ -128,7 +130,7 @@ const MainInfoPage = ({
               </Form.Control.Feedback>
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-year">
               <Form.Label>
                 Published Year
                 {' '}
@@ -150,7 +152,7 @@ const MainInfoPage = ({
               </Form.Control>
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-author">
               <Form.Label>
                 {' '}
                 Authors
@@ -159,6 +161,7 @@ const MainInfoPage = ({
               </Form.Label>
               {renderAuthors(values, touched, errors, handleChange, setValues)}
               <Button
+                id="publication-add-author"
                 className="mt-2"
                 variant="secondary"
                 onClick={() => {
@@ -169,7 +172,7 @@ const MainInfoPage = ({
               </Button>
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-desc">
               <Form.Label>
                 Description
                 {' '}
@@ -190,7 +193,7 @@ const MainInfoPage = ({
               </Form.Control.Feedback>
             </Form.Group>
 
-            <Form.Group>
+            <Form.Group id="publication-link">
               <Form.Label>Link</Form.Label>
               <Form.Control
                 className="placeholder-text"
--- a/client/src/components/publications/form/ProfileLinkPage.js
+++ b/client/src/components/publications/form/ProfileLinkPage.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import { Formik } from 'formik';
 import * as yup from 'yup';
 import { useDispatch, useSelector } from 'react-redux';
@@ -10,6 +11,7 @@ import { importPublications } from '../../../actions/publications';
 import { IMPORT_FAIL, UPDATE_GSCHOLAR_ID } from '../../../actions/types';
 import '../publications.css';
 import { PrimaryButton, DangerButton } from '../../shared/styledComponents';
+import { PublicationsImportFormWalkthrough } from '../publicationsOnboarding';
 
 const ProfileLinkPage = ({ closeModal }) => {
   const teamId = useSelector((state) => state.team.teamId);
@@ -60,7 +62,8 @@ const ProfileLinkPage = ({ closeModal }) => {
         handleSubmit, handleChange, values, touched, errors,
       }) => (
         <Form noValidate onSubmit={handleSubmit}>
-          <Form.Group>
+          <PublicationsImportFormWalkthrough />
+          <Form.Group id="import-publication-link">
             <Form.Label>
               Enter your Google Scholar profile link below:
             </Form.Label>
--- a/client/src/components/publications/publication/Publication.js
+++ b/client/src/components/publications/publication/Publication.js
@@ -22,6 +22,7 @@ import '../publications.css';
 import {
   SecondaryButton, DangerButton, StyledButtonGroup, ButtonGroupItem,
 } from '../../shared/styledComponents';
+import { PublicationsEditDeleteWalkthrough } from '../publicationsOnboarding';
 
 const Publication = ({ pub }) => {
   const dispatch = useDispatch();
@@ -137,6 +138,7 @@ const Publication = ({ pub }) => {
 
   return (
     <div
+      id="publication-card"
       className="publication-container mb-2"
       onMouseOver={handleMouseOver}
       onFocus={handleMouseOver}
@@ -151,7 +153,7 @@ const Publication = ({ pub }) => {
           <Col md={10}>
             <div style={{ display: 'flex' }}>
               <div style={{ paddingTop: '10px', paddingLeft: '10px' }}>
-                <input type="checkbox" checked={checkedPublications.includes(pub._id) || false} />
+                <input id="publication-checkbox" type="checkbox" checked={checkedPublications.includes(pub._id) || false} />
               </div>
               <div className="pubs-title">
                 {pub.link ? (
@@ -163,6 +165,9 @@ const Publication = ({ pub }) => {
                 ) : (
                   pub.title
                 )}
+              </div>
+              <div style={{ paddingLeft: '10px' }}>
+                <PublicationsEditDeleteWalkthrough />
               </div>
             </div>
           </Col>
--- a/client/src/components/publications/publicationsLayout/PublicationsEditor.js
+++ b/client/src/components/publications/publicationsLayout/PublicationsEditor.js
@@ -82,7 +82,10 @@ const PublicationsEditor = ({
           <div style={{ display: 'flex' }}>
 
             <Dropdown>
-              <StyledDropdownToggle style={{ whiteSpace: 'normal' }}>
+              <StyledDropdownToggle
+                id="add-publication-button"
+                style={{ whiteSpace: 'normal' }}
+              >
                 Add Publication(s)
               </StyledDropdownToggle>
               <Dropdown.Menu>
@@ -104,6 +107,7 @@ const PublicationsEditor = ({
             >
               <div style={{ display: 'inline-block', cursor: 'not-allowed' }}>
                 <DangerButton
+                  id="delete-publication-icon"
                   style={{ marginLeft: '5px' }}
                   onClick={() => setShowDeleteMessage(true)}
                   disabled={checkedPublications.length === 0}
@@ -122,7 +126,9 @@ const PublicationsEditor = ({
 
         <Col md={6}>
           <div style={{ display: 'flex' }}>
-            <StyledButtonGroup>
+            <StyledButtonGroup
+              id="publication-group-by"
+            >
               Group By
               {' '}
               {Object.keys(groupByOptions).map((groupBy) => (
@@ -136,7 +142,9 @@ const PublicationsEditor = ({
                 </ButtonGroupItem>
               ))}
             </StyledButtonGroup>
-            <StyledButtonGroup>
+            <StyledButtonGroup
+              id="publication-sort-by"
+            >
               Sort By
               {' '}
               {Object.keys(sortingOptions).map((sortBy) => (
@@ -175,6 +183,7 @@ const PublicationsEditor = ({
         <Col md={2}>
           <OverlayTrigger trigger={['hover', 'focus']} placement="bottom" overlay={ButtonHint}>
             <PrimaryButton
+              id="publication-update-layout"
               className="float-right"
               onClick={handleUpdate}
             >
--- a/client/src/components/publications/publicationsOnboarding.js
+++ b/client/src/components/publications/publicationsOnboarding.js
@@ -1,0 +1,263 @@
+/**
+ * The Steps component displays the different pop-ups for each onboarding message in sequence
+ */
+
+import React, { useState } from 'react';
+import { Steps } from 'intro.js-react';
+import { BiHelpCircle } from 'react-icons/bi';
+import 'intro.js/introjs.css';
+
+const PublicationsPageWalkthrough = () => {
+  const [intro, setIntro] = useState({
+    stepsEnabled: false,
+    initialStep: 0,
+    steps: [
+      {
+        element: '#add-publication-button',
+        title: 'Add Publication(s)',
+        intro: 'Click here to add a new publication manually or by importing from Google Scholar',
+      },
+      {
+        element: '#delete-publication-icon',
+        title: 'Deleting selected publications',
+        intro: 'Click here to delete all selected publications',
+      },
+      {
+        element: '#publication-group-by',
+        title: 'Grouping Publications',
+        intro: 'Click here to group all publications by category (or none)',
+      },
+      {
+        element: '#publication-sort-by',
+        title: 'Sorting Publications',
+        intro: 'Click here to sort all publications by title, author or year',
+      },
+      {
+        element: '#publication-update-layout',
+        title: 'Update Layout',
+        intro: 'Click here to update the layout of publications in the deployed site.',
+      },
+    ],
+  });
+
+  const toggleSteps = () => {
+    setIntro({ ...intro, stepsEnabled: !intro.stepsEnabled });
+  };
+
+  const onExit = () => {
+    setIntro({ ...intro, stepsEnabled: false });
+  };
+  return (
+    <div>
+      <Steps
+        enabled={intro.stepsEnabled}
+        steps={intro.steps}
+        initialStep={intro.initialStep}
+        onExit={onExit}
+      />
+      <BiHelpCircle onClick={toggleSteps} />
+    </div>
+  );
+};
+
+const PublicationsEditDeleteWalkthrough = () => {
+  const [intro, setIntro] = useState({
+    stepsEnabled: false,
+    initialStep: 0,
+    steps: [
+      {
+        element: '#publication-checkbox',
+        intro: 'Click on this checkbox to select the added publication.',
+        position: 'bottom',
+      },
+      {
+        element: '#publication-card',
+        intro: 'Hover over the card to edit or delete an publication. Icons will appear in the top right corner of the publication header.',
+        position: 'right',
+      },
+    ],
+  });
+
+  const toggleSteps = () => {
+    setIntro({ ...intro, stepsEnabled: !intro.stepsEnabled });
+  };
+
+  const onExit = () => {
+    setIntro({ ...intro, stepsEnabled: false });
+  };
+
+  return (
+    <div>
+      <Steps
+        enabled={intro.stepsEnabled}
+        steps={intro.steps}
+        initialStep={intro.initialStep}
+        onExit={onExit}
+      />
+      <BiHelpCircle onClick={toggleSteps} />
+    </div>
+  );
+};
+
+const PublicationsManualFormWalkthrough = () => {
+  const [intro, setIntro] = useState({
+    stepsEnabled: false,
+    initialStep: 0,
+    steps: [
+      {
+        element: '#publication-title-form',
+        title: 'New Publication',
+        intro: 'Please ensure that all the mandatory fields are filled. These are marked by an asterisk next to them.',
+      },
+      {
+        element: '#publication-title',
+        title: 'Publication Title',
+        intro: 'Please ensure that the title is 3-60 characters',
+      },
+      {
+        element: '#publication-year',
+        title: 'Published Year',
+        intro: 'Use the drop down menu to select the year.',
+      },
+      {
+        element: '#publication-author',
+        title: 'Publication Author(s)',
+        intro: 'Please enter the authors of this publication',
+      },
+      {
+        element: '#publication-add-author',
+        title: 'Add Author',
+        intro: 'Click here to add an author to this publication',
+      },
+      {
+        element: '#publication-desc',
+        title: 'Publication Description',
+        intro: 'Enter the description of the publication(5 - 500 characters)',
+      },
+      {
+        element: '#publication-link',
+        title: 'Publication Link',
+        intro: 'Please ensure a valid link to this publication is provided',
+      },
+    ],
+  });
+
+  const toggleSteps = () => {
+    setIntro({ ...intro, stepsEnabled: !intro.stepsEnabled });
+  };
+
+  const onExit = () => {
+    setIntro({ ...intro, stepsEnabled: false });
+  };
+
+  return (
+    <div>
+      <Steps
+        enabled={intro.stepsEnabled}
+        steps={intro.steps}
+        initialStep={intro.initialStep}
+        onExit={onExit}
+      />
+      <BiHelpCircle onClick={toggleSteps} />
+    </div>
+  );
+};
+
+const PublicationsManualCategoryFormWalkthrough = () => {
+  const [intro, setIntro] = useState({
+    stepsEnabled: false,
+    initialStep: 0,
+    steps: [
+      {
+        element: '#publication-category',
+        title: 'New Publication',
+        intro: 'Select the category of your publication',
+      },
+      {
+        element: '#publication-category-title',
+        title: 'Publication Category Title',
+        intro: 'Please ensure that the title is 3-60 characters',
+      },
+      {
+        element: '#publication-volume',
+        title: 'Published Volume',
+        intro: 'Please ensure that the volume of the publication is a number',
+      },
+      {
+        element: '#publication-issue',
+        title: 'Publication Issue',
+        intro: 'Please ensure that the issue of the publication is a number',
+      },
+      {
+        element: '#publication-pages',
+        title: 'Publication Pages',
+        intro: 'Please ensure that the pages of the publication is a number',
+      },
+      {
+        element: '#publication-publisher',
+        title: 'Publication Publisher',
+        intro: 'Enter the publisher of the publication',
+      },
+    ],
+  });
+
+  const toggleSteps = () => {
+    setIntro({ ...intro, stepsEnabled: !intro.stepsEnabled });
+  };
+
+  const onExit = () => {
+    setIntro({ ...intro, stepsEnabled: false });
+  };
+
+  return (
+    <div>
+      <Steps
+        enabled={intro.stepsEnabled}
+        steps={intro.steps}
+        initialStep={intro.initialStep}
+        onExit={onExit}
+      />
+      <BiHelpCircle onClick={toggleSteps} />
+    </div>
+  );
+};
+
+const PublicationsImportFormWalkthrough = () => {
+  const [intro, setIntro] = useState({
+    stepsEnabled: false,
+    initialStep: 0,
+    steps: [
+      {
+        element: '#import-publication-link',
+        title: 'Import Publication',
+        intro: 'Import your publication(s) by providing a valid Google Scholar link',
+      },
+    ],
+  });
+
+  const toggleSteps = () => {
+    setIntro({ ...intro, stepsEnabled: !intro.stepsEnabled });
+  };
+
+  const onExit = () => {
+    setIntro({ ...intro, stepsEnabled: false });
+  };
+
+  return (
+    <div>
+      <Steps
+        enabled={intro.stepsEnabled}
+        steps={intro.steps}
+        initialStep={intro.initialStep}
+        onExit={onExit}
+      />
+      <BiHelpCircle onClick={toggleSteps} />
+    </div>
+  );
+};
+
+export default PublicationsPageWalkthrough;
+export { PublicationsManualFormWalkthrough };
+export { PublicationsEditDeleteWalkthrough };
+export { PublicationsManualCategoryFormWalkthrough };
+export { PublicationsImportFormWalkthrough };


cyr17:
The static analyzer is failing because of localStorage being an undefined variable, this can be fixed by including the localstorage as a global variable in the .eslintrc

aksh0001 (changes requested):
Please configure your linter to adhere to the existing indenting scheme. There are 700 changes in the `team.js` controller and it's hard to review the actual changes.

client/src/components/profileInfoEdit/ProfileDeleteModal.js
```diff
@@ -0,0 +1,62 @@
...
+      toast.error('Log in with github account')
+    } 
+    else{
+      try{
+      dispatch(deleteGHPages(teamId,access_token))
+      dispatch(deleteTeam(teamId))
+      dispatch(logout())
+      } catch(error){
+        toast.error('GitHub Pages doesnt exist')
+      }
```
aksh0001, line 30: Interesting that the SA didn't catch these indents and the missing semicolons. Thoughts @vgoh1 ?

aksh0001: We'll need to fix the SA if that's the case. The style is very inconsistent here.

aksh0001: If it's conflicting with `prettier` rules, which may be more flexible, we may need to change prettier instead.

vgoh1: Yeah, let me look into why the linter isn't catching the formatting + the prettier issue. I personally want to make prettier format according to airbnb because it's faster doing the shortcut to format than running npm run lint.

vgoh1: @cyr17 Can you run `npm run lint` in the /api and /client and fix the errors that come up? They come up there for me. The formatting stuff can be fixed using `npm run lint -- --fix`.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -49,12 +48,20 @@ const ProfileInfoEdit = () => {
...
+
+  // const updateProfile = (teamId, profileData) => {
+  //   try {
+  //     dispatch(updateTeam(teamId, profileData))
+  //     toast.success('Profile has been successfully updated');
+  //   } catch (error) {
+  //     console.error(error);
+  //     toast.error('Profile has not been updated');
+  //   }
+  // };
```
rpra0006, line 61: Maybe comments can be taken out?

client/src/components/profileInfoEdit/ProfileDeleteModal.js
```diff
@@ -0,0 +1,62 @@
...
+     <Modal show={deleteAlert}>
+       <Modal.Header className="modalHeader">
+         <Modal.Title> Delete Account! </Modal.Title>
+       </Modal.Header>
+       <Modal.Body>
+         Are you sure you want to delete your account? All significant data will be deleted!
+       </Modal.Body>
+       <Modal.Footer>
+       <Button variant="light" onClick={() => setdeleteAlert(false)}>
+           back
```
rpra0006, line 48: "back" can be capitalised for consistency

client/src/components/profileInfoEdit/ProfileDeleteModal.js
```diff
@@ -0,0 +1,62 @@
...
+         </Button>
+         <Button variant="danger" onClick={partialDelete}>
+           Keep github page
```
rpra0006, line 51: Same here with capitalization "gihub page"

rpra0006: Either the buttons layout or color should change. Hard to distinguish delete all data and keep github page button. (For accessibility)

rpra0006 (changes requested):

client/src/components/profileInfoEdit/ProfileDeleteModal.js
vgoh1: Also my 2c, but I think the two buttons should be on the profile page rather than in this modal, this modal should just be  for confirming whether you want to delete your account. 
Deleting the account means everything needs to go, the data in the DB + repo. 
Deleting the data should mean the data in the DB (except for team doc since that's the account) + repo as well imo because would you want to keep the data in the repo if you're starting from a clean state?
It doesn't make sense to keep the GH pages site if you're leaving the site, or if you want to start the site again.
Anybody, feel free to add your own thoughts to the discussion.

client/.eslintrc.json
```diff
@@ -4,10 +4,14 @@
         "es2021": true,
         "node": true
     },
+    "globals": {
+        "localStorage": true
+    },
```
vgoh1, line 9: Didn't know you could do this! Good catch, could you also refactor the other places that disable this error on a line by line basis?

api/src/routes/team.js
```diff
@@ -74,4 +74,20 @@ teamRouter.patch(
...
+  teamMiddleware.validateTeamId,
+  teamController.deleteTeam
+);
+
+teamRouter.delete(
+  '/:team_id/deleteGHPages/:code',
+  authMiddleware.cookieJwtAuth,
+  mongooseMiddleware.validateTeamObjectId,
+  teamMiddleware.validateTeamId,
+  teamController.deleteGHPages
```
vgoh1, line 90: Also, the backend functionality is missing from this PR?

client/src/components/profileInfoEdit/ProfileDeleteModal.js
cyr17: So do you reckon have Clear all data and Delete account as the two buttons instead?

cyr17: Imo we thought we'd have two options,
delete all data + repo and delete website data but keep the repo (not for restarting )
but yeah this can be easily refactored to the functionalities above!

client/src/components/profileInfoEdit/ProfileInfoEdit.js
cyr17: I got this from a previous pull but yeah will surely take this out!

client/src/components/profileInfoEdit/ProfileDeleteModal.js
cyr17: This is wierd xp,
npm run lint on /api
image

npm run lint on /client
image

cyr17:
> Please configure your linter to adhere to the existing indenting scheme. There are 700 changes in the `team.js` controller and it's hard to review the actual changes.

Im pretty confused on how to tackle this but for reviewing the actual changes you can view the conflicting files section, ill try to fix that but as ive said before im a bit confused on how to do it .

api/src/routes/team.js
cyr17: wait what?

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function deleteTeam(req, res, next) {
+  try {
+    const { team_id: _id } = req.params;
+
+    await HomePage.deleteOne({ teamId: _id });
+    await Website.deleteOne({ teamId: _id });
+    await Achievement.deleteOne({ teamId: _id });
+    await Publication.deleteOne({ teamId: _id });
```
vgoh1, line 333: iirc this only deletes the first publication it finds, you'll need to use deleteMany()

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+    await Team.findByIdAndDelete(_id);
+
+    res.status(200).json('deleted successfully');
```
vgoh1, line 336: Message should start on a capital

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+async function clearTeam(req, res, next) {
+  try {
+    const { team_id: _id } = req.params;
+
+    await HomePage.deleteOne({ teamId: _id });
+    await Website.deleteOne({ teamId: _id });
+    await Achievement.deleteOne({ teamId: _id });
+    await Publication.deleteOne({ teamId: _id });
+
+    res.status(200).json('cleared successfully');
```
vgoh1, line 361: Message should start on a capital

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function clearTeam(req, res, next) {
+  try {
+    const { team_id: _id } = req.params;
+
+    await HomePage.deleteOne({ teamId: _id });
+    await Website.deleteOne({ teamId: _id });
+    await Achievement.deleteOne({ teamId: _id });
+    await Publication.deleteOne({ teamId: _id });
```
vgoh1, line 359: Use deleteMany() here, same as comment above

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+async function deleteGHPages(req, res, next) {
+  const { code } = req.params;
+  // Call github API to get username
+  const { data } = await axios.get('https://api.github.com/user',
+    {
+      headers: { Authorization: `token ${code}` },
+    });
+  if (data.errors) {
+    return next(
+      fillErrorObject(400, 'Validation error', [data.errors[0].detail]),
```
vgoh1, line 386: Is it a validation error here or something else?

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+  // Creating repoName
+  const ghUsername = data.login;
+  logger.info(`GitHub deploy initiated for user: ${ghUsername}`);
+  const repoName = `${ghUsername}.github.io`;
+
+  // delete repo
+
+  let deletePages;
+  try {
+    deletePages = await axios({
```
vgoh1, line 399: weird the linter didn't call this out, but you should define the var in L399 with const instead to follow convention

vgoh1, line 399: you can also do an axios.delete() for readability

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+          repoName
+        }`,
+      method: 'delete',
+      headers: {
+        Authorization: `token ${code}`,
+        Accept: 'application/vnd.github.v3+json',
+      },
+    });
+    // result logged
+    res.status(200).json('deleted successfully');
```
vgoh1, line 413: Fix up the grammar/punctuation for this message

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+/**
+  * Delete the team from the database on /team/:team_id
+  * @param {} req request object, containing team id in the url
+  * @param {*} res response object, the deleted team document
+  * @returns 200: teamn is deleted
+  * @returns 404: team is not found
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function deleteGHPages(req, res, next) {
+  const { code } = req.params;
```
vgoh1, line 378: what is code? do you mean the token? the variable name should be self-explanatory

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+        Authorization: `token ${code}`,
+        Accept: 'application/vnd.github.v3+json',
+      },
+    });
+    // result logged
+    res.status(200).json('deleted successfully');
+    if (deletePages.status === 204) {
+      logger.info(`GitHub pages deleted for user: ${ghUsername}`);
+    } else {
+      logger.info(`GitHub pages deleted for user: ${ghUsername} failed!`);
```
vgoh1, line 417: shouldn't this throw an error if it failed to delete the repo?

api/src/controllers/team.js
vgoh1: also you're deleting the repo + pages so a better name for this var would be `deleteRepo` or something along those lines

api/src/routes/team.js
```diff
@@ -74,4 +74,28 @@ teamRouter.patch(
...
+teamRouter.delete(
+  '/:team_id/clear',
+  authMiddleware.cookieJwtAuth,
+  mongooseMiddleware.validateTeamObjectId,
+  teamMiddleware.validateTeamId,
+  teamController.clearTeam,
+);
+
+teamRouter.delete(
+  '/:team_id/deleteGHPages/:code',
```
vgoh1, line 94: if code is meant to be the token, probably don't pass it in the URL, pass it in the body instead

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+
+/**
+  * Clear the team Data from the database on /team/:team_id
+  * @param {} req request object, containing team id in the url
+  * @param {*} res response object, the deleted team document
+  * @returns 200: teamn is deleted
+  * @returns 404: team is not found
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function clearTeam(req, res, next) {
```
vgoh1, line 352: The only difference between this and deleteTeam is that the latter deletes the team doc, right? So you can combine the endpoints into one and just have a toggle on whether you need to delete the team doc or not to save on duplicated code.

client/.eslintrc.json
```diff
@@ -4,11 +4,16 @@
...
     "extends": ["eslint:recommended", "airbnb"],
     "parserOptions": {
         "ecmaVersion": 12
     },
+    
     "rules": {
+        "linebreak-style": 0,
```
vgoh1, line 16: what's the purpose of suppressing this rule?

client/src/components/profileInfoEdit/ProfileClearModal.js
```diff
@@ -0,0 +1,48 @@
...
+import { clearTeam, deleteGHPages } from '../../actions/team';
+
+const ProfileClearModal = ({ clearAlert, setclearAlert }) => {
+  const { teamId } = useSelector((state) => state.team);
+  const dispatch = useDispatch();
+
+  const HandleClear = () => {
+    const access_token = localStorage.getItem('GH_access_token');
+    if (access_token === null) {
+      toast.error('Log in with github account');
```
vgoh1, line 18: Instead of doing it like this, if the user wants to delete their website/repo, it'd be nice to let them log in on this page and not have to go to the dashboard to login and then come back to the profile page. So you can take the same logic as the github button and use it in this flow.

client/src/components/profileInfoEdit/ProfileClearModal.js
```diff
@@ -0,0 +1,48 @@
...
+import { Button, Modal } from 'react-bootstrap';
+import { useSelector, useDispatch } from 'react-redux';
+import { clearTeam, deleteGHPages } from '../../actions/team';
+
+const ProfileClearModal = ({ clearAlert, setclearAlert }) => {
+  const { teamId } = useSelector((state) => state.team);
+  const dispatch = useDispatch();
+
+  const HandleClear = () => {
+    const access_token = localStorage.getItem('GH_access_token');
```
vgoh1, line 16: Try to avoid camel case here, I know it's suppressed in the eslint rules but its pending refactor so if we use snake case here we can avoid more work for that eventual refactor. :)

client/src/components/profileInfoEdit/ProfileClearModal.js
```diff
@@ -0,0 +1,48 @@
...
+    if (access_token === null) {
+      toast.error('Log in with github account');
+    } else {
+      try {
+        dispatch(deleteGHPages(teamId, access_token));
+      } catch (error) {
+        toast.error('GitHub Pages doesnt exist');
+      }
+      dispatch(clearTeam(teamId));
+      toast.success('profile data has been successfully cleared');
```
vgoh1, line 26: I think it's better to show a success toast within the dispatch action instead of using a generic toast here

client/src/actions/team.js
```diff
@@ -309,3 +351,30 @@ export const updateTeamTheme = (teamId, themeData) => async (dispatch) => {
...
+ * This action creator will be called when a user want to delete their deployed GitHub Pages
+ *
+ * @param {*} teamId id of the team
+ * @param {*} teamData data object of the data to be deleted
+ * @returns
+ */
+export const deleteGHPages = (teamId, accessToken) => async (dispatch) => {
+  try {
+    dispatch({
+      type: DELETE_REQUEST,
```
vgoh1, line 365: I don't see this being used anywhere in the frontend?

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+  if (data.errors) {
+    return next(
+      fillErrorObject(400, 'Validation error', [data.errors[0].detail]),
+    );
+  }
+
+  // Creating repoName
+  const ghUsername = data.login;
+  logger.info(`GitHub deploy initiated for user: ${ghUsername}`);
+  const repoName = `${ghUsername}.github.io`;
```
vgoh1, line 393: what happens if this repo doesn't exist? should you validate that first?

api/src/controllers/team.js
```diff
@@ -310,6 +315,114 @@ async function updateTeam(req, res, next) {
...
+    });
+  if (data.errors) {
+    return next(
+      fillErrorObject(400, 'Validation error', [data.errors[0].detail]),
+    );
+  }
+
+  // Creating repoName
+  const ghUsername = data.login;
+  logger.info(`GitHub deploy initiated for user: ${ghUsername}`);
```
vgoh1, line 392: update the message here

client/src/components/profileInfoEdit/ProfileDeleteModal.js
```diff
@@ -0,0 +1,49 @@
...
+ * Delete Alert Modal component display a modal and ask for confirmation when the delete button is clicked
+ */
+import React from 'react';
+import toast from 'react-hot-toast';
+import { Button, Modal } from 'react-bootstrap';
+import { useSelector, useDispatch } from 'react-redux';
+import { logout } from '../../actions/auth';
+import { deleteTeam, deleteGHPages } from '../../actions/team';
+
+const ProfileDeleteModal = ({ deleteAlert, setdeleteAlert }) => {
```
vgoh1, line 11: use camel case for setdeleteAlert -> `setDeleteAlert`

client/.eslintrc.json
cyr17: Getting lint to print expect LF linebreaks instead of CRLF , and by default all of our IDLES use LF so having this supresses that error!

api/src/controllers/team.js
cyr17: Yea the token is used to validate / retrieve the username right?
And btw this was copied from the deployPages() ,for context

api/src/controllers/team.js
cyr17: Yes it does and if an error is thrown there is a catch block following this throwing 500, Error occurred within the server!

api/src/controllers/team.js
vgoh1: Ah right, I remember now, I had it as validation error because if the call failed it would mean the token was bad. Maybe it's worth changing the message to something more descriptive related to GH.

client/src/components/profileInfoEdit/ProfileClearModal.js
cyr17: which variables are you reffereing to tho? ihve had the clearAlert etc in camel case to follow the consistency but yeah can you point out the variable you are trying to refer to?

api/src/controllers/team.js
vgoh1: Which means it'll never hit the else block if it fails. So it's more appropriate for the logging statement to be somewhere else, also, because of the res.status(200) in L413, L414 - 415 never actually get run.

client/src/components/profileInfoEdit/ProfileClearModal.js
vgoh1: Sorry, I meant avoid snake case LOL (don't know what I was thinking when I wrote that)
`access_token` should be `accessToken`

client/.eslintrc.json
vgoh1: @aksh0001 can weigh in on whether we keep this rule or not

client/src/components/profileInfoEdit/ProfileClearModal.js
cyr17: Right xp

client/.eslintrc.json
cyr17: I mean it does no harm, maybe try having windows users test the commit before this fix and see if they have the same lint issue?

cyr17: This is the issue I was referring to!
https://stackoverflow.com/questions/37826449/expected-linebreaks-to-be-lf-but-found-crlf-linebreak-style

api/src/controllers/team.js
cyr17: done

api/src/controllers/team.js
```diff
@@ -310,6 +315,102 @@ async function updateTeam(req, res, next) {
...
+  * @returns 200: teamn is deleted
+  * @returns 404: team is not found
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function clearTeam(req, res, next) {
+  try {
+    const { team_id: _id } = req.params;
+    const { isDeleteFlag } = req.body;
+    await HomePage.deleteMany({ teamId: _id });
+    await Website.deleteMany({ teamId: _id });
```
vgoh1, line 331: Just an FYI that each team should only have one homepage and website doc so you don't actually need to use deleteMany() here but that's not an issue. :)

api/src/controllers/team.js
```diff
@@ -310,6 +315,102 @@ async function updateTeam(req, res, next) {
...
+      headers: {
+        Authorization: `token ${ghToken}`,
+        Accept: 'application/vnd.github.v3+json',
+      },
+    });
+    if (repoValidator.status === 200) validFlag = true;
+    else validFlag = false;
+  } catch (error) {
+    validFlag = false;
+  }
```
vgoh1, line 382: I suggest making this a middleware so that this function's logic doesn't get too long

vgoh1: Because if it doesn't exist, it shouldn't even try to delete it right? So it can return early

api/src/controllers/team.js
```diff
@@ -310,6 +315,102 @@ async function updateTeam(req, res, next) {
...
+      logger.info(` Failed: GitHub pages not deleted for user: ${ghUsername}`);
+      return next(
+        fillErrorObject(500, 'Error occurred with server', [error.message]),
+      );
+    }
+  } else {
+    logger.info(` Failed: GitHub pages does not exist for user: ${ghUsername}`);
+  }
+  return next(
+    fillErrorObject(500, 'Github Pages doesnt exist for the user', ['Github Pages doesnt exist for the user']),
```
vgoh1, line 410: punctuation in the message, missing apostrophe

client/.eslintrc.json
vgoh1, line 16: nitpick: instead of 0, use off so its consistent with the other rules

client/src/actions/types.js
```diff
@@ -36,6 +38,9 @@ export const GET_GH_ACCESS_TOKEN = 'GET_GH_ACCESS_TOKEN';
 export const DEPLOY_REQUEST = 'DEPLOY_REQUEST';
 export const DEPLOY_SUCCESS = 'DEPLOY_SUCCESS';
 export const DEPLOY_FAIL = 'DEPLOY_FAIL';
+export const DELETE_REQUEST = 'DELETE_REQUEST';
```
vgoh1, line 41: if this isn't being used, can be removed

client/src/components/profileInfoEdit/ProfileClearModal.js
```diff
@@ -0,0 +1,47 @@
...
+const ProfileClearModal = ({ clearAlert, setClearAlert }) => {
+  const { teamId } = useSelector((state) => state.team);
+  const dispatch = useDispatch();
+
+  const HandleClear = () => {
+    const accessToken = localStorage.getItem('GH_access_token');
+    try {
+      dispatch(deleteGHPages(teamId, accessToken));
+    } catch (error) {
+      toast.error('GitHub Pages doesnt exist');
```
vgoh1, line 21: The action should be using an error message toast to say the same thing, so you don't need a toast here?

client/src/components/profileInfoEdit/ProfileDeleteModal.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * Delete Alert Modal component display a modal and ask for confirmation when the delete button is clicked
+ */
```
vgoh1, line 3: This file has almost the same code with ProfileClearModal.js, can you refactor it so that it's in the same file?

api/src/controllers/team.js
```diff
@@ -310,6 +315,102 @@ async function updateTeam(req, res, next) {
...
 
+/**
+  * Clear the team Data from the database on /team/:team_id
+  * @param {} req request object, containing team id in the url
+  * @param {*} res response object, the deleted team document
+  * @returns 200: teamn is deleted
+  * @returns 404: team is not found
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function clearTeam(req, res, next) {
```
vgoh1, line 326: I'm not a huge fan of calling this `clearTeam` because it's not clear to me (pun unintended) straight away what it's trying to do. Any thoughts on a better name? `resetTeamData`?

client/src/actions/team.js
```diff
@@ -286,6 +285,24 @@ export const updateTeam = (teamId, teamData) => async (dispatch) => {
...
+ * @param {*} teamId id of the team
+ * @param {*} teamData data object of the data to be deleted
+ * @returns
+ */
+export const clearTeam = (teamId, isDelete) => async (dispatch) => {
+  try {
+    const body = {
+      isDeleteFlag: isDelete,
+    };
+    await api.clearTeam(teamId, body);
```
vgoh1, line 300: Should there be a success dispatch after here?

client/src/components/profileInfoEdit/ProfileClearModal.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+    const isDeleteFlag = false;
+    dispatch(clearTeam(teamId, isDeleteFlag));
+    setClearAlert(false);
+    dispatch(successMessageCreator('Profile data cleared successfully!'));
```
vgoh1, line 26: Prefer this being in the actions rather than in the components.

client/src/actions/team.js
vgoh1: Also remember to merge main into your branch and deal with the conflicts in this file

api/src/controllers/team.js
```diff
@@ -310,6 +315,102 @@ async function updateTeam(req, res, next) {
...
+  try {
+    const { team_id: _id } = req.params;
+    const { isDeleteFlag } = req.body;
+    await HomePage.deleteMany({ teamId: _id });
+    await Website.deleteMany({ teamId: _id });
+    await Achievement.deleteMany({ teamId: _id });
+    await Publication.deleteMany({ teamId: _id });
+    if (isDeleteFlag === true) {
```
vgoh1, line 334: if you're checking a boolean you can just do `if (isDeleteFlag)`

api/src/controllers/team.js
```diff
@@ -310,6 +315,102 @@ async function updateTeam(req, res, next) {
...
+        Authorization: `token ${ghToken}`,
+        Accept: 'application/vnd.github.v3+json',
+      },
+    });
+    if (repoValidator.status === 200) validFlag = true;
+    else validFlag = false;
+  } catch (error) {
+    validFlag = false;
+  }
+  if (validFlag === true) {
```
vgoh1, line 383: if you're checking a boolean you can just do `if (validFlag)`

api/src/controllers/team.js
```diff
@@ -310,6 +315,102 @@ async function updateTeam(req, res, next) {
...
+  * @param {*} res response object, the deleted team document
+  * @returns 200: teamn is deleted
+  * @returns 404: team is not found
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function clearTeam(req, res, next) {
+  try {
+    const { team_id: _id } = req.params;
+    const { isDeleteFlag } = req.body;
+    await HomePage.deleteMany({ teamId: _id });
```
pmyip3, line 330: After I click clear data, I got an error when refreshing the page. I think it's because the website document of that team had been deleted, but getWebsiteInfo api is called each time when the page is refreshed. 
If only the 'Clear Data' button is clicked (i.e. isDeleteFlag != true), I think the homepage and website shouldn't be deleted, instead, maybe just reset the properties inside these documents?

api/src/controllers/team.js
```diff
@@ -16,14 +21,14 @@ const {
...
- * @returns 500: error trying to update the document in db
- */
+  * Associates a twitter handle with a team on the /team/twitter-handle/:team-id endpoint.
+  * @param {*} req request object, containing the team_id in the url and twitter handle in the body
+  * @param {*} res response object
+  * @returns 200: successful added twitter handle to team
+  * @returns 400: team id is not in a valid hexadecimal format
+  * @returns 404: team is not found, or handle is invalid
+  * @returns 500: error trying to update the document in db
+  */
```
aksh0001, line 31: nit: maybe don't add a space between the jsdocs here. It'll inflate the diff count and I believe most of our IDE's are configured to format the `*` in the old way. Ditto for other unnecessary spaces.

cyr17: this was from the eslint fix btw!

client/src/actions/team.js
cyr17: line 300 uses the successmessage generator which does the toast

api/src/controllers/team.js
```diff
@@ -68,11 +71,11 @@ async function storeHandle(req, res, next) {
...
- * Gets the team info
- * @param {*} req request object contains the teamId decoded in auth middleware
- * @param {*} res response object, the team related info
- * @returns 200: the team related info
- */
+  * Gets the team info
+  * @param {*} req request object contains the teamId decoded in auth middleware
+  * @param {*} res response object, the team related info
+  * @returns 200: the team related info
+  */
```
aksh0001, line 78: please fix spaces, ditto for other docs

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -74,99 +76,121 @@ const ProfileInfoEdit = () => {
...
-          </div>
-          <div className="my-1">
-            <Button variant="danger" onClick={profileDeleted}>
-              Delete account
-            </Button>
-          </div>
-        </Form>
-      </Container>
-    </div>
+    <>
```
rpra0006, line 90: What is this empty html tag for?

api/src/controllers/team.js
cyr17: the thing is in the actual file it just has one space each but in the git preview it shows two spaces, can you confirm if its 1 spaced by opening it on VS code?

client/src/components/profileInfoEdit/ProfileInfoEdit.js
cyr17: So I mainly followed what we had before and from what I understand it is a JSX fragment.  And when I remove it I get this error :
 Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>?

api/src/middleware/team.js
```diff
@@ -23,6 +24,45 @@ async function validateTeamId(req, res, next) {
...
+      );
+    }
+    next();
+  } catch (error) {
+    next(
+      fillErrorObject(404, 'GH pages not found!', [
+        'GitHub Repo doesnt exist for this team!',
+      ]),
+    );
+  }
```
vgoh1, line 63: What's the difference between these two? Shouldn't the next() in the catch block be a different error like server error?

api/src/middleware/team.js
```diff
@@ -23,6 +24,45 @@ async function validateTeamId(req, res, next) {
   next();
 }
 
+async function validateTeamRepo(req, res, next) {
+  const { ghToken } = req.body;
+  const { data } = await axios.get('https://api.github.com/user',
+    {
+      headers: { Authorization: `token ${ghToken}` },
+    });
```
vgoh1, line 32: What happens if this errors out (eg invalid token)? Do you need a try/catch block here to handle it?

client/src/actions/team.js
```diff
@@ -293,6 +295,34 @@ export const updateTeam = (teamId, teamData) => async (dispatch) => {
...
+ */
+export const resetTeamData = (teamId, isDelete) => async (dispatch) => {
+  try {
+    const body = {
+      isDeleteFlag: isDelete,
+    };
+    await api.resetTeamData(teamId, body);
+    dispatch({
+      type: DELETE_TEAM_PUBLICATIONS,
+      payload: null,
```
vgoh1, line 313: can remove this line if you're not returning anything

client/src/actions/team.js
```diff
@@ -293,6 +295,34 @@ export const updateTeam = (teamId, teamData) => async (dispatch) => {
...
+    });
+    dispatch({
+      type: RESET_TEAM,
+      payload: null,
```
vgoh1, line 317: can remove this line if you're not returning anything

api/src/controllers/team.js
```diff
@@ -313,6 +316,75 @@ async function updateTeam(req, res, next) {
...
+  * @returns 404: team is not found
+  * @returns 400: team id is not in a valid hexadecimal format
+  */
+async function deleteGHPages(req, res, next) {
+  const { ghToken } = req.body;
+  // Call github API to get username
+  const { data } = await axios.get('https://api.github.com/user',
+    {
+      headers: { Authorization: `token ${ghToken}` },
+    });
```
vgoh1, line 358: You need a try/block here because you don't validate the token before this call.
1. Does the token exist in localStorage? This can be done on the frontend, this way it won't reach this call if there is no token.
2. If the token has expired/is not valid, this will fail, hence you need a try/catch here.

client/src/actions/website.js
```diff
@@ -65,6 +66,18 @@ export const deletePage = (teamId, pageName) => async (dispatch) => {
   }
 };
 
+export const resetWebPage = (teamId) => async (dispatch) => {
+  try {
+    api.resetWebPage(teamId);
+    dispatch({
+      type: RESET_WEBPAGE,
+      payload: null,
```
vgoh1, line 74: can remove this line if no payload

client/src/components/profileInfoEdit/ProfileResetModal.js
```diff
@@ -0,0 +1,68 @@
...
+  const { teamId } = useSelector((state) => state.team);
+  const dispatch = useDispatch();
+
+  const isDeleteFlag = type;
+  const titleMessage = isDeleteFlag ? 'Delete Account Data!' : 'Clear Account Data!';
+  const warningMessage = isDeleteFlag ? 'Are you sure you want to delete your account? ' : 'Are you sure you want to reset your account? ';
+  const buttonName = isDeleteFlag ? 'Delete' : 'Clear';
+
+  const HandleResetType = () => {
+    const accessToken = localStorage.getItem('GH_access_token');
```
vgoh1, line 24: What happens if the access token is not found in localStorage? See my comment earlier: https://github.com/Researchify/Researchify/pull/198/files#r711673343

api/src/controllers/team.js
vgoh1: Also because we're seeing this block of logic being duplicated, might be good to pull this up into the middleware.

api/src/middleware/team.js
```diff
@@ -23,6 +24,45 @@ async function validateTeamId(req, res, next) {
...
+  if (data.errors) {
+    next(
+      fillErrorObject(400, 'Validation error: user doesnt exist!', [data.errors[0].detail]),
+    );
+  }
```
vgoh1, line 37: Pull this up into its own function in this middleware since it's being used somewhere else. See related comment: https://github.com/Researchify/Researchify/pull/198/files#r711673608

api/src/controllers/team.js
```diff
@@ -313,6 +316,75 @@ async function updateTeam(req, res, next) {
...
+    );
+    if (deleteRepo.status === 204) {
+      logger.info(`GitHub pages deleted for user: ${ghUsername}`);
+    }
+    // result logged
+    return res.status(200).json('Deleted successfully!');
+  } catch (error) {
+    logger.info(` Failed: GitHub pages not deleted for user: ${ghUsername}!`);
+    return next(
+      fillErrorObject(500, 'Error occurred with server!', [error.message]),
```
vgoh1, line 383: The error message that shows up cause of this line is: `Request failed with status code 403` and might not be too user friendly. Might be worth changing to something like 'Failed to delete GitHub pages'.

api/src/controllers/team.js
cyr17: 1:  This is being handled in the front end and, if so the ghLoginModal is called and prompts the user to log in!
2: If the token is not valid the validateTeamRepo from the middleware catches this and exits early ,the code doesn't reach this function
3: This particular line is being duplicated to retrieve the username from the ghToken , the solution to this would be sending the username and reponame as part of the body , but for this i would have to make an api call from the client /front end.

api/src/middleware/team.js
cyr17: The same as before it just retrieves the username from the token and here it exits if it is not valid!

client/src/components/profileInfoEdit/ProfileResetModal.js
cyr17: IF the token is not found it is being handled in the front end and the GhLoginModal gets triggered which prompts the user to log in with their github

api/src/middleware/team.js
cyr17: Good catch!

client/src/components/profileInfoEdit/GhLogInModal.js
```diff
@@ -0,0 +1,69 @@
...
+    >
+      <GoMarkGithub className="mr-2" />
+      Login with GitHub
+    </GitHubLogin>
+  );
+
+  return (
+    <Modal show={logInAlert}>
+      <Modal.Header className="modalHeader">
+        <Modal.Title> Log in! </Modal.Title>
```
vgoh1, line 47: can we change this to `Please login to GitHub`

client/src/components/profileInfoEdit/GhLogInModal.js
```diff
@@ -0,0 +1,69 @@
...
+      </Modal.Header>
+      <Modal.Body>
+        Looks like you need to log in to continue the ongoing process!
+
+        Please Login!
```
vgoh1, line 52: `You will need to login to your GitHub account to proceed with clearing your account data` might be better? It also begs the question - if they never deployed but want to reset their data, they might be wondering why do I need to login to github? Should we be explaining why?

api/src/middleware/team.js
```diff
@@ -12,15 +13,61 @@ async function validateTeamId(req, res, next) {
...
+      );
+    }
+    return next();
+  } catch (error) {
+    return next(
+      fillErrorObject(500, 'Server Error!', [
+        'Failed to access the repository!',
+      ]),
+    );
+  }
```
vgoh1, line 70: When I clear data without the repo being made, it goes into the catch block so it's not going through L56-62. Might want to double check this logic.

client/src/actions/team.js
```diff
@@ -316,3 +344,22 @@ export const updateTeamTheme = (teamId, themeData) => async (dispatch) => {
...
+ */
+export const deleteGHPages = (teamId, accessToken) => async (dispatch) => {
+  try {
+    const body = {
+      ghToken: accessToken,
+    };
+    await api.deleteGHPages(teamId, body);
+    dispatch(successMessageCreator('The deployed GH Pages have been deleted'));
+  } catch (err) {
+    dispatch(errorActionGlobalCreator(err));
```
vgoh1, line 363: Related to this comment: https://github.com/Researchify/Researchify/pull/198/files#r711858215
Assuming the user just wanted to reset their data, would they get alarmed at the error toast that says github repo not found?

client/src/actions/team.js
```diff
@@ -316,3 +344,22 @@ export const updateTeamTheme = (teamId, themeData) => async (dispatch) => {
...
+ * @param {*} teamData data object of the data to be deleted
+ * @returns
+ */
+export const deleteGHPages = (teamId, accessToken) => async (dispatch) => {
+  try {
+    const body = {
+      ghToken: accessToken,
+    };
+    await api.deleteGHPages(teamId, body);
+    dispatch(successMessageCreator('The deployed GH Pages have been deleted'));
```
vgoh1, line 361: nit: change the message to `Your GitHub Pages website has been deleted.`

client/src/actions/team.js
```diff
@@ -293,6 +295,32 @@ export const updateTeam = (teamId, teamData) => async (dispatch) => {
...
+    };
+    await api.resetTeamData(teamId, body);
+    dispatch({
+      type: DELETE_TEAM_PUBLICATIONS,
+    });
+    dispatch({
+      type: RESET_TEAM,
+    });
+
+    dispatch(successMessageCreator('Profile has been cleared!'));
```
vgoh1, line 318: nit: message can be `Your account data has been reset.` or something

client/src/components/profileInfoEdit/ProfileResetModal.js
```diff
@@ -0,0 +1,68 @@
...
+import { resetTeamData, deleteGHPages } from '../../actions/team';
+import { resetHomepage } from '../../actions/homepage';
+import { resetWebPage } from '../../actions/website';
+
+const ProfileResetModal = ({ resetAlert, setResetAlert, type }) => {
+  const { teamId } = useSelector((state) => state.team);
+  const dispatch = useDispatch();
+
+  const isDeleteFlag = type;
+  const titleMessage = isDeleteFlag ? 'Delete Account Data!' : 'Reset Account Data!';
```
vgoh1, line 19: change to `Delete Account` and `Reset Account Data`

client/src/components/profileInfoEdit/ProfileResetModal.js
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <Modal show={resetAlert}>
+      <Modal.Header className="modalHeader">
+        <Modal.Title>
+          {titleMessage}
+        </Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        {warningMessage}
+        All significant data will be deleted!
```
vgoh1, line 48: `All the data associated with your account will be deleted along with your account itself.`?

vgoh1:
Found a bug - If I go to reset data, and click back on the login modal, this modal shows up:
image
Same thing happens if I click on delete data.

api/src/controllers/team.js
```diff
@@ -313,6 +316,68 @@ async function updateTeam(req, res, next) {
...
+async function resetTeamData(req, res, next) {
+  try {
+    const { teamId } = req.params;
+    const { isDeleteFlag } = req.body;
+    await Achievement.deleteMany({ teamId });
+    await Publication.deleteMany({ teamId });
+    if (isDeleteFlag) {
+      await Team.findByIdAndDelete(teamId);
+      return res.status(200).json('Deleted successfully!');
+    } return res.status(200).json('Cleared successfully!');
```
vgoh1, line 336: full stops instead of exclamation marks please

rpra0006:
When I click on "Reset Data" the publications, team data, etc. is removed from sidebar. But when I add the respective pages I still get the original data. Was this intentional?

cyr17:
> When I click on "Reset Data" the publications, team data, etc. is removed from sidebar. But when I add the respective pages I still get the original data. Was this intentional?
> 
> Delete account functionality works for me though 👍

Yes , this should be like this but can you check if the publications is removed though ( it should be ).

client/src/components/profileInfoEdit/GhLogInModal.js
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <Modal show={logInAlert}>
+      <Modal.Header className="modalHeader">
+        <Modal.Title> Please Log in to GitHub </Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        You will need to login to your GitHub account to proceed with clearing your account data
+        The github repo would not be deleted if you do not login.
+        Please Login!
```
vgoh1, line 52: How about this: You will need to login to your GitHub account to proceed with clearing your account data. However, if you choose to not login, your GitHub repository containing your website data will not be deleted.

api/src/controllers/team.js
```diff
@@ -313,6 +316,68 @@ async function updateTeam(req, res, next) {
...
+ * @returns 200: teamn is deleted
+ * @returns 404: team is not found
+ * @returns 400: team id is not in a valid hexadecimal format
+ */
+async function resetTeamData(req, res, next) {
+  try {
+    const { teamId } = req.params;
+    const { isDeleteFlag } = req.body;
+    await Achievement.deleteMany({ teamId });
+    await Publication.deleteMany({ teamId });
```
pmyip3, line 332: I think team members from the Team schema needed to be reset to the empty array as well. I still got the previous team members list after I reset data and then adding back the team page.

client/src/components/profileInfoEdit/ProfileResetModal.js
```diff
@@ -0,0 +1,70 @@
...
+import { resetTeamMember, resetTeamData, deleteGHPages } from '../../actions/team';
+import { resetHomepage } from '../../actions/homepage';
+import { resetWebPage } from '../../actions/website';
+
+const ProfileResetModal = ({ resetAlert, setResetAlert, type }) => {
+  const { teamId } = useSelector((state) => state.team);
+  const dispatch = useDispatch();
+
+  const isDeleteFlag = type;
+  const titleMessage = isDeleteFlag ? 'Delete Account!' : 'Reset Account Data!';
```
vgoh1, line 19: remove exclamation mark

vgoh1:
I think the reset data/delete account buttons are triggering the submit form logic: 
image
Because the 'team is updated' toast shows up after I click them

client/src/components/profileInfoEdit/ProfileResetModal.js
```diff
@@ -0,0 +1,70 @@
...
+
+const ProfileResetModal = ({ resetAlert, setResetAlert, type }) => {
+  const { teamId } = useSelector((state) => state.team);
+  const dispatch = useDispatch();
+
+  const isDeleteFlag = type;
+  const titleMessage = isDeleteFlag ? 'Delete Account' : 'Reset Account Data';
+  const warningMessage = isDeleteFlag ? 'Are you sure you want to delete your account? ' : 'Are you sure you want to reset your account? ';
+  const buttonName = isDeleteFlag ? 'Delete' : 'Reset';
+  const warningContent = isDeleteFlag ? 'All significant data will be reset!' : 'All significant data will be deleted including your account data!';
```
vgoh1, line 22: I think the text should be swapped: 
image
When you're deleting the account, the second line should show rather than the first?

vgoh1 (approved):

api/src/controllers/team.js
```diff
@@ -313,6 +336,68 @@ async function updateTeam(req, res, next) {
...
+  try {
+    const deleteRepo = await axios.delete(
+      `https://api.github.com/repos/${ghUsername}/${repoName}`,
+      {
+        headers: {
+          Authorization: `token ${ghToken}`,
+          Accept: 'application/vnd.github.v3+json',
+        },
+      },
+    );
```
aksh0001, line 387: Prefer using `octokit`, see https://github.com/Researchify/Researchify/pull/250

api/src/routes/team.js
```diff
@@ -54,6 +54,14 @@ teamRouter.patch(
...
 );
 
+teamRouter.delete(
+  '/:teamId/reset-members',
+  authMiddleware.cookieJwtAuth,
+  mongooseMiddleware.validateTeamObjectId,
+  teamMiddleware.validateTeamId,
+  teamController.resetTeamMember,
+);
+
```
aksh0001, line 64: any reason why this is a `delete` endpoint while the other is a `post`?

api/src/controllers/team.js
```diff
@@ -313,6 +336,68 @@ async function updateTeam(req, res, next) {
...
+ * @param {*} res response object, the deleted team document
+ * @returns 200: teamn is deleted
+ * @returns 404: team is not found
+ * @returns 400: team id is not in a valid hexadecimal format
+ */
+async function resetTeamData(req, res, next) {
+  try {
+    const { teamId } = req.params;
+    const { isDeleteFlag } = req.body;
+    await Achievement.deleteMany({ teamId });
```
aksh0001, line 351: What about deleting the other models? Homepage, theme, etc …

aksh0001: Tbh I think you should make this functionality as a database trigger rather than encapsulate all this logic in the controller. I can help with that if you need

api/src/controllers/team.js
```diff
@@ -313,6 +336,68 @@ async function updateTeam(req, res, next) {
...
+          Authorization: `token ${ghToken}`,
+          Accept: 'application/vnd.github.v3+json',
+        },
+      },
+    );
+    if (deleteRepo.status === 204) {
+      logger.info(`GitHub pages deleted for user: ${ghUsername}`);
+    }
+    // result logged
+    return res.status(200).json('Deleted successfully!');
```
aksh0001, line 392: This should be an error level log

api/src/controllers/team.js
```diff
@@ -183,6 +186,26 @@ function updateTeamMember(req, res, next) {
...
 
+/**
+ * Reset the team member from the database on /team/:teamId/member.
+ * @param {*} req request object, containing team id in the url
+ * @param {*} res response object, the updated team member document
+ * @returns 200: the team member was updated
+ * @returns 404: team is not found
+ * @returns 400: team id is not in a valid hexadecimal format
+ */
+async function resetTeamMember(req, res, next) {
```
aksh0001, line 197: Curious, why an endpoint to specifically remove team members ? Shouldn’t team member removal be part of the broader delete as well?

api/src/controllers/website.js
```diff
@@ -81,6 +81,45 @@ async function deleteWebPage(req, res, next) {
  * @returns 200: changes updated in DB
  * @returns 500: Server error while saving to DB
  */
+async function updatePublicationOptions(req, res, next) { // eslint-disable-line no-unused-vars
```
aksh0001, line 84: Is this part of delete functionality?

api/src/controllers/website.js
```diff
@@ -81,6 +81,45 @@ async function deleteWebPage(req, res, next) {
...
+        },
+      },
+    );
+    return res.status(200).json(updatedPubOptions);
+  } catch (err) {
+    return res.send(fillErrorObject(500, 'Server error', [err.errors]));
+  }
+}
+
+async function resetWebPage(req, res, next) { // eslint-disable-line no-unused-vars
```
aksh0001, line 102: Similar Q, why am endpoint specifically for this?
I’d imagine this to be part of broader delete as well?

api/src/middleware/team.js
```diff
@@ -12,15 +13,61 @@ async function validateTeamId(req, res, next) {
...
   }
 
   req.foundTeam = foundTeam; // todo: does this need to be set inside middleware?
-  next();
+  return next();
+}
+
+async function validateGhUser(req, res, next) {
+  const { ghToken } = req.body;
+  const { data } = await axios.get('https://api.github.com/user',
```
aksh0001, line 29: Use octokit to validate existence

aksh0001: Ditto for other raw requests

client/src/actions/team.js
```diff
@@ -298,3 +319,75 @@ export const updateTeam = (teamId, teamData) => async (dispatch) => {
...
+    });
+
+    dispatch(successMessageCreator('Your account data has been reset.'));
+  } catch (error) {
+    dispatch(errorActionGlobalCreator(error));
+  }
+};
+
+/**
+ * This action creater find/create a new theme and update it in team data.
```
aksh0001, line 353: Typos

client/src/actions/team.js
```diff
@@ -298,3 +319,75 @@ export const updateTeam = (teamId, teamData) => async (dispatch) => {
...
+ * @param {*} teamId
+ * @param {*} themeData
+ * @returns
```
aksh0001, line 356: Please don’t add a docstring and leave fields empty. You should either populate them or remove the docs entirely

aksh0001 (changes requested):
Sorry but I’m a little confused. According to the PR desc. I’d imagine there two be at most two endpoints: one for deleting an entire teams data (clean wipe) and one to delete their data but retain their deployed website. Could you explain the reasoning behind adding more loosely tied functionality in this PR, like resetting webpages, etc?

api/src/controllers/website.js
cyr17: Nope

client/src/actions/team.js
cyr17: This was from the merge, i get weird stuff like this on my laptop ig

api/src/controllers/team.js
cyr17: > Sorry but I’m a little confused. According to the PR desc. I’d imagine there’d be at most two endpoints: one for deleting an entire teams data (clean wipe) and one to delete their data but retain their deployed website. Could you explain the reasoning behind adding more loosely tied functionality in this PR, like resetting webpages, etc?

So it was initally exactly this , but I refactored the functionality based on the reviews i got throughout PI3

package-lock.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "researchify",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
aksh0001, line 4: revert this file too

rsal0007:
> Works well. Produces no errors on the console. Rename the back button on the modal to "Continue without Github" or something like that as it may be a concern from the UX perspective.

+1

rsal0007 (approved):

package-lock.json
aksh0001: @cyr17 please revert before merging

client/src/reducers/index.js
```diff
@@ -38,7 +38,7 @@ const appReducer = combineReducers({
  */
 const rootReducer = (state, action) => {
   // Reset state when we log out or if the user has reset their data.
-  if (action.type === LOG_OUT || action.type === RESET_TEAM_DATA) {
+  if (action.type === LOG_OUT) {
```
aksh0001, line 41: fix the comment above

client/src/reducers/authReducer.js
```diff
@@ -28,6 +28,8 @@ const authReducer = (state = INITIAL_AUTH_STATE, action) => {
       return { ...state, loading: false };
     case LOG_OUT:
       return { ...state, loading: false, logIn: false };
+    case RESET_TEAM_DATA:
+      return INITIAL_AUTH_STATE;
```
aksh0001, line 32: should this be just return state?

cyr17: return INITIAL_STATE as its the default for undefined state , which we were using before

PR #202, opened 2021-09-04
Make orgName to be the title on client website

pmyip3:
# Description 
As title

# Type of change 
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
Currently, the title on the deployed website is "React App", which is not nice 

# Solution description
The title would change based on the pages the user is browsing - implemented by using a third-party library, react helmet
      e.g. Publications - orgName

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

image
image

Deploy to  github page to see if the custom title is shown on the website 

https://www.npmjs.com/package/react-helmet

# Risk
Low.


api/src/controllers/website.js
```diff
@@ -151,9 +151,28 @@ async function updatePublicationOptions(req, res, next) { // eslint-disable-line
   }
 }
 
+async function updateTitle(req, res, next) { // eslint-disable-line no-unused-vars
```
aksh0001, line 154: Docs.

client/src/actions/website.js
```diff
@@ -76,3 +77,16 @@ export const updatePublicationOptions = (teamId, preference) => async (dispatch)
...
 };
+
+export const updateWebsiteTitle = (teamId, website) => async (dispatch) => {
+  try {
+    await api.updateWebsiteTitle(teamId, website);
+    dispatch({
+      type: UPDATE_WEBSITE_TITLE,
+      payload: website.websiteTitle,
+    });
+    dispatch(successMessageCreator('Title had been updated'));
```
aksh0001, line 88: nit: typo `Title has been updated.`

scholly/base/src/components/App.js
```diff
@@ -19,6 +21,9 @@ const App = () => {
 
   return (
     <>
+      <Helmet>
+        <title>{WEB_PAGES.title}</title>
```
aksh0001, line 25: Maybe best to follow the convention where we're assigning the global to a local variable inside the component before accessing it (to avoid any potential modifications). Could you make a local  inside this component and access `.title` from it, similar to how we're doing in other components?

scholly/base/src/components/landingPage/LandingPage.js
```diff
@@ -3,12 +3,22 @@
...
+import { TEAM_HOMEPAGE, WEB_PAGES } from '../../global/data';
 
 const landingPage = () => {
   const homepageData = TEAM_HOMEPAGE;
   return (
     <>
+      <Helmet>
+        <title>
+          {' '}
+          {WEB_PAGES.title}
```
aksh0001, line 16: Ditto.

scholly/base/src/components/publications/PublicationPage.js
```diff
@@ -43,6 +45,15 @@ const PublicationPage = () => (
   // </Fragment>
 
   <>
+    <Helmet>
+      <title>
+        {' '}
+        {WEB_PAGES.title}
```
aksh0001, line 51: Here, too.

scholly/base/src/components/team/TeamPage.js
```diff
@@ -3,13 +3,23 @@
...
 import TeamMember from './TeamMember';
 
 const TeamPage = () => {
   const teamMembers = TEAM_MEMBERS;
   return (
     <>
+      <Helmet>
+        <title>
+          {' '}
+          {WEB_PAGES.title}
```
aksh0001, line 17: And here.

api/src/controllers/website.js
```diff
@@ -151,9 +151,28 @@ async function updatePublicationOptions(req, res, next) { // eslint-disable-line
...
+  const { team_id: _id } = req.params;
+  try {
+    await Website.updateOne(
+      { teamId: _id },
+      {
+        $set: {
+          title: updatedTitle,
+        },
+      },
+    );
```
kdav108, line 165: What happens if a team doesn't originally have a title (and so `title` is not an attribute stored under website schema on mongoDB), and they add it later. Does `$set` still work?

api/src/models/website.model.js
```diff
@@ -20,6 +20,8 @@ const websiteSchema = new mongoose.Schema(
     title: {
       type: String,
       required: false,
+      minlength: 3,
+      maxlength: 30,
```
kdav108, line 24: Just curious to know if these numbers are set randomly or derived from somewhere?

scholly/base/src/components/App.js
```diff
@@ -3,6 +3,8 @@
  */
 import React from 'react';
 import { Switch, Route } from 'react-router-dom';
+import { Helmet } from 'react-helmet';
```
kdav108, line 6: Nice never new this existed. It makes updating the title much simpler than I had initially thought.

kdav108 (commented):
Great job, works like a charm! 

Just a UI/UX suggestion regarding the deploy form:
Screen Shot 2021-09-05 at 10 51 23 am

How about we merge the `Save Title` button with the `Deploy to GitHub Pages` button? Since users can only enter the title when deploying, and right after using the `Save Title` button they're almost guaranteed to click deploy, we can reduce the number of clicks required from 2 to just one, if we only have one button to do both jobs.

If we want we can name the combined button `Save and Deploy`, since the title of the from gives away that we're deploying to GitHub Pages anyways, we don't need it repeated on the button title. 

Again, it's just a small suggestion, happy to hear everyone's views.

api/src/models/website.model.js
pmyip3: I set this restriction and it's consistent with the validation in the front-end program. Let me know if the length does not make sense.

api/src/controllers/website.js
pmyip3: yup, it will still work.

pmyip3:
> Great job, works like a charm!
> 
> Just a UI/UX suggestion regarding the deploy form:
> Screen Shot 2021-09-05 at 10 51 23 am
> 
> How about we merge the `Save Title` button with the `Deploy to GitHub Pages` button? Since users can only enter the title when deploying, and right after using the `Save Title` button they're almost guaranteed to click deploy, we can reduce the number of clicks required from 2 to just one, if we only have one button to do both jobs.
> If we want we can name the combined button `Save and Deploy`.
> 
> Again, it's just a small suggestion, happy to hear everyone's views.

I was thinking to implement that at first too, but there are two separate API calls so I don't want to mix them together in the same function. I am happy to change that. Any other thoughts?

api/src/controllers/website.js
```diff
@@ -99,9 +99,35 @@ async function updatePublicationOptions(req, res, next) { // eslint-disable-line
   }
 }
 
+/**
+ * Update website title on the deployed website
```
aksh0001, line 103: Could you elaborate on what title we're deploying here, it's not very clear from just reading the docs.

Maybe include a line saying `The title is the HTML <title> tag used for client websites. Currently, we are using the team's name for the title, but this route can be used to update title if need be`/similar.

client/src/components/dashboard/deploy/form/DeployForm.js
```diff
@@ -0,0 +1,88 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import {
+  Row, InputGroup, Button, Form, Col,
+} from 'react-bootstrap';
+
+import { updateWebsiteTitle } from '../../../../actions/website';
+import { deployToGHPages } from '../../../../actions/team';
+
+const DeployForm = ({ teamId }) => {
+  const { title } = useSelector((state) => state.website);
```
aksh0001, line 12: I thought we were going to simply embed the team's name for the title in scholly. Any reason that didn't work out? If we're sticking with that approach, I feel like we shouldn't expose code that will not be used. Ditto for the corresponding functionality, including actions, etc.

aksh0001: If we're using team name as title, it makes sense to leave in the API functionality in case that requirement arises in the future. But IMO we shouldn't expose UI to allow users to edit it.

scholly/base/src/components/team/TeamPage.js
```diff
@@ -3,13 +3,24 @@
...
 import React, { Fragment } from 'react';
 import { CardDeck, Container } from 'react-bootstrap';
-import { TEAM_MEMBERS } from '../../global/data';
+import { Helmet } from 'react-helmet';
+import { TEAM_MEMBERS, WEB_PAGES } from '../../global/data';
 import TeamMember from './TeamMember';
 
 const TeamPage = () => {
   const teamMembers = TEAM_MEMBERS;
+  const { title } = WEB_PAGES;
```
aksh0001, line 12: A custom title is overkill IMO, if we know the team name, the title can just be:

`Team Page - <Team Name>`. That makes it more meaningful than a custom title I reckon.

scholly/base/src/components/publications/PublicationPage.js
```diff
@@ -41,14 +43,25 @@ const PublicationPage = () => (
...
-  <>
-    <Container className="pages-top-padding text-center mt-3 mb-3">
-      <div className="publication-pg-title">Our Publications</div>
-    </Container>
-    <Container fluid>
-      <Publications />
-    </Container>
-  </>
-);
+  const { title } = WEB_PAGES;
```
aksh0001, line 46: `Publications - <Team Name>` could be used as the title.

scholly/base/src/components/landingPage/LandingPage.js
```diff
@@ -3,12 +3,23 @@
  */
 import React from 'react';
 import { Container } from 'react-bootstrap';
-import { TEAM_HOMEPAGE } from '../../global/data';
+import { Helmet } from 'react-helmet';
+import { TEAM_HOMEPAGE, WEB_PAGES } from '../../global/data';
 
 const landingPage = () => {
   const homepageData = TEAM_HOMEPAGE;
+  const { title } = WEB_PAGES;
```
aksh0001, line 11: `Home - <Team Name>` could be used as the title.

client/src/components/dashboard/deploy/form/DeployForm.js
pmyip3: Oh, I thought we are going to let the custom title default to the orgName (see lines 13 and 24) and the user has the choice to modify it?! 
So you mean we just left the deploy UI the same on Researchicy as before and just grab the orgName in scholly to be the title?

aksh0001: Yep that's what I was thinking. Unless we think that there is value in setting the title.
@vgoh1 @kdav108 @rpra0006 @jfan0010

rpra0006: I don't see much value in customizing the browser tab title, imo having Team/Org Name by default should be enough.

jfan0010: Personally agree with Akshay and Richard, but I don't think Chunyang will give in to this. If you check his GitHub page, he's also customizing the tab title himself.
image

aksh0001: Makes sense, only thing with that is that Chunyang's page is a personal reflection of his own research career. Since Researchify is centered around Research **teams** do we still see value in the custom title?

kdav108: I feel given that this is implemented already, there's little weight in rejecting all the effort now. Why not just keep the team name as default and if any researcher (like Chunyang) wants to personalise for SEO reason or otherwise, they can still do it. We can let the user decide.

Also agree with Akshay's comment that Chunyang's argument is more reasonable for personal sites and not exactly team pages.

rpra0006: Yes we can keep the functionality but at a UI standpoint isn't it a bit confusing for less technical people when they enter a "Website Title"?. I'd expect the website title to be what people see (actual site) when they visit my website and not the browser tab title.

aksh0001: +1. We definitely don't want to diminish anyone's effort, but I do believe our UI will be better off without it. IMO, it's an anti-pattern to leave in functionality that doesn't serve justifiable value.

aksh0001 (changes requested):
Great job on the PR !  It works perfectly fine :100:  Personally, I feel like we should discard the UI that allows clients to edit their title for discoverability and other UX reasons. We can leave the controller functionality in though.

jfan0010 (approved):
Works on mine!

Little opinion: Would setting `teamName` (eg. Allan Lab) as website title better than `orgName` (eg. Monash) ?



ps. Maybe we can defer this PR a bit if and only if PR#200 is merging soon

aksh0001 (approved):
LGTM, approval.

client/src/components/dashboard/deploy/form/DeployForm.js
vgoh1: Researchify is centered around Research teams <- there's a chance that solo users will be using it right? Not necessarily just teams.
I'm with Kunj, for SEO reasons as well. And I do agree with allowing customization of the browser tab title be confusing, it could be an 'advanced' option that's hidden away unless the user clicks on it? We can discuss it in our winter meeting on how to handle it properly on Sunday.

vgoh1 (approved):

pmyip3:
> Works on mine!
> 
> Little opinion: Would setting `teamName` (eg. Allan Lab) as website title better than `orgName` (eg. Monash) ?

Yes, I think it does make more sense. Thanks for the suggestion :)

cyr17:
> > Great job, works like a charm!
> > Just a UI/UX suggestion regarding the deploy form:
> > Screen Shot 2021-09-05 at 10 51 23 am
> > How about we merge the `Save Title` button with the `Deploy to GitHub Pages` button? Since users can only enter the title when deploying, and right after using the `Save Title` button they're almost guaranteed to click deploy, we can reduce the number of clicks required from 2 to just one, if we only have one button to do both jobs.
> > If we want we can name the combined button `Save and Deploy`.
> > Again, it's just a small suggestion, happy to hear everyone's views.
> 
> I was thinking to implement that at first too, but there are two separate API calls so I don't want to mix them together in the same function. I am happy to change that. Any other thoughts?

Say you had to rechange your title how would you do that if its in a single UI button?

PR #204, opened 2021-09-06
ProfileInfoEdit.js upload picture functionality

rpra0006:
# Description 
Upload file in ProfileInfoEdit.js now updates the front end profile icon. Added string type to store image in team model, to be continued in the backend.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [X] Refactoring
- [ ] Documentation 

# Problem
Upload image doesn't do anything currently.

# Solution description
Image targets source to file uploaded.

# Tests 
All unit tests pass?
Y
Builds successfully?
Y

# Attach screenshot of the expected result (e.g. UI of the website/ postman api call result/ database changes)
image

# Provide steps in detail for reviewers to reproduce the expected result of the code changes.

# Risk
Low


api/src/models/team.model.js
```diff
@@ -63,6 +63,10 @@ const teamSchema = new mongoose.Schema(
       type: mongoose.SchemaTypes.ObjectId,
       ref: 'theme',
     },
+    teamImage: {
+      type: String,
+      required: false,
+    },
```
kdav108, line 69: A nice have: check that the input is base64. (Not required if it takes too much time)

kdav108 (dismissed):
Nice quick PR! Great work!

Is the default picture empty only for me?
Screen Shot 2021-09-06 at 8 54 29 pm

rpra0006:
> Nice quick PR! Great work!
> 
> Is the default picture empty only for me?
> Screen Shot 2021-09-06 at 8 54 29 pm

Yeah it's not fetching anything at the moment. So we'll probably have to set a default pic in the backend logic.

api/src/models/team.model.js
rpra0006: I don't know where to put this in the backend 🤣 . There is a input check for image MIME types in the file upload button itself though.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -67,12 +84,12 @@ const ProfileInfoEdit = () => {
...
           <Form.Group controlId="formProfilePic">
             <Image
-              src={profilePic}
+              ref={uploadedImage}
               roundedCircle
               height="184px"
               width="184px"
             />
-            <Form.Control type="file" />
+            <Form.Control type="file" accept="image/*" onChange={handleImageUpload} multiple="false" />
```
kdav108, line 92: Sorry for some reason this comment was left out (I blame GitHub)
Currently It does allow selecting multiple files:
Screen Shot 2021-09-06 at 9 00 13 pm

We can change it to `multiple={false}` to fix it, I think it expects a boolean instead of a string. Also, fyi a string is treated as `true` in JS, and that's why it's allowing multiple files.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -20,6 +20,8 @@ import { updateTeam } from '../../actions/team';
 const ProfileInfoEdit = () => {
   const dispatch = useDispatch();
 
+  const uploadedImage = React.useRef(profilePic); // To be referred to server default profile image
```
pmyip3, line 23: Although the backend is not implemented, would this variable be retrieved from the redux state cause the team image is one of the properties in the teamInfo object?

pmyip3 (commented):
> > Nice quick PR! Great work!
> > Is the default picture empty only for me?
> > Screen Shot 2021-09-06 at 8 54 29 pm
> 
> Yeah it's not fetching anything at the moment. So we'll probably have to set a default pic in the backend logic.

A follow-up question, as the default pic is not working anyway, can the `import profilePic from '../../images/profilepic.jpg';` on line14 be moved and just set null on line 23? `const uploadedImage = useRef(null);`

kdav108:
> > > Nice quick PR! Great work!
> > > Is the default picture empty only for me?
> > > Screen Shot 2021-09-06 at 8 54 29 pm
> > 
> > 
> > Yeah it's not fetching anything at the moment. So we'll probably have to set a default pic in the backend logic.
> 
> A follow-up question, as the default pic is not working and it will be fetched from the backend, can the `import profilePic from '../../images/profilepic.jpg';` on line14 be moved and just set null on line 23? `const uploadedImage = useRef(null);`

I feel like the default pic shouldn't come from backend, we can just use the pic we currently have `../../images/profilepic.jpg;`. This way we avoid having to make an extra get request to backend to get the default pic.

api/src/models/team.model.js
aksh0001: nit: prefer renaming this to something more meaningful like `profilePic`/similar

api/src/models/team.model.js
```diff
@@ -63,6 +63,10 @@ const teamSchema = new mongoose.Schema(
       type: mongoose.SchemaTypes.ObjectId,
       ref: 'theme',
     },
+    teamImage: {
```
aksh0001, line 66: Also, do we need to implement corresponding API functionality to make this PR complete?

client/.eslintrc.json
```diff
@@ -32,6 +33,7 @@
         "no-param-reassign": "off", // should refactor later
         "no-return-assign": "off", // should refactor later
         "spaced-comment": "off",
-        "arrow-body-style": "off" // should refactor later
+        "arrow-body-style": "off", // should refactor later
+        "linebreak-style": "off"
```
aksh0001, line 37: If we're ignoring this rule, does this mean that when a Mac/Linux dev uses different line breaks, it registers as a diff, even though the code may not have changed?

client/src/components/profileInfoEdit/ProfileInfoEdit.js
aksh0001: +1, we should implement the corresponding Redux functionality in this PR.

api/src/models/team.model.js
rpra0006: Was thinking of opening another PR for api functionality to keep this PR small.

client/.eslintrc.json
rpra0006: Saw this in a discussion from @cyr17 @vgoh1 in PR#198. So I wasn't sure if there's a better approach to surpressing linebreak style.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: > > > > Nice quick PR! Great work!
> > > > Is the default picture empty only for me?
> > > > Screen Shot 2021-09-06 at 8 54 29 pm
> > > 
> > > 
> > > Yeah it's not fetching anything at the moment. So we'll probably have to set a default pic in the backend logic.
> > 
> > 
> > A follow-up question, as the default pic is not working and it will be fetched from the backend, can the `import profilePic from '../../images/profilepic.jpg';` on line14 be moved and just set null on line 23? `const uploadedImage = useRef(null);`
> 
> I feel like the default pic shouldn't come from backend, we can just use the pic we currently have `../../images/profilepic.jpg;`. This way we avoid having to make an extra get request to backend to get the default pic.

Sure that's alright then, I just added a simple if statement to check if image is undefined from redux state.

rpra0006: Empty profilePic redux state:
image

Preview of image upload:
image

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: Thanks that works

client/.eslintrc.json
vgoh1: I don't think it's being registered as diff? Because we have people using different OS and pushing to the repo and that hasn't been an issue. According to [eslint](https://eslint.org/docs/rules/linebreak-style): `Many versioning systems (like git and subversion) can automatically ensure the correct ending. However to cover all contingencies, you can activate this rule.` so I don't think we have to worry about this rule going in, since git should fix it for us.

rpra0006:
image
Ignore the profileImg, that was a previous update I made. Anyways, it is stored as base64 binary string in mongodb in profilePic.

client/.eslintrc.json
aksh0001: Makes sense, resolving.

api/src/controllers/publications.js
```diff
@@ -4,10 +4,12 @@
  */
 const mongoose = require('mongoose');
 const logger = require('winston');
-const { firefox } = require('playwright');
+const axios = require('axios');
+const cheerio = require('cheerio');
+const iconv = require('iconv-lite');
```
jfan0010, line 9: A dumb question regarding Git, aren't the changes in this file (and other publication files) already in main? Why is it still showing in Files Changed?

rpra0006: I just noticed this 🤣 . Was it because I merged main into this branch?

rpra0006: Update - Merging normally seems to do the trick, the more you know I guess..

vgoh1: Yeah, I've done a squash merge from main into my branch before and cause it checks the diffs via the commit hash, everything is in a new commit hash after you squash and thinks its new. :/ Alas.

vgoh1:
While testing, I run into this error:
image
When I click upload again, and then cancel, I get this:
image
This doesn't seem to happen when you start from no pic and then cancel, so you might need to handle this case specifically.
Also is it possible to make the select a file box pop up when they click the button? Because if you click the filename text next to it, it also opens the popup.

jfan0010 (commented):
Since there are only 3 files changed, do you wanna include passing it into redux in this PR? ( ͡° ͜ʖ ͡°)
I think we can leave the API in the next PR if that's too big.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -21,22 +21,49 @@ const ProfileInfoEdit = () => {
...
 
   const {
-    teamId, teamName, orgName, email,
+    teamId, teamName, orgName, email, profilePic,
   } = useSelector(
     (state) => state.team,
   );
 
-  const [profileData, setInputs] = useState({ teamName, orgName, email });
+  const [profileData, setInputs] = useState({
```
aksh0001, line 29: nit: rename the setter to `setProfileData` to be consistent with React `useState` naming conventions.

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -67,12 +94,12 @@ const ProfileInfoEdit = () => {
 
           <Form.Group controlId="formProfilePic">
             <Image
-              src={profilePic}
+              src={profileData.profilePic}
```
aksh0001, line 97: minor: maybe we can use the `Formik` validation style we're using for other components since we're using a Form here? @pmyip3 @mable-yip @jfan0010 @jxiangfan

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -21,22 +21,49 @@ const ProfileInfoEdit = () => {
...
   };
 
+  /**
+   * Updates profile image field when user uploads file
+  */
+
+  // If profilePic is undefined, set a default profile pic
+  if (!profileData.profilePic) {
+    profileData.profilePic = defaultProfilePic;
+  }
```
aksh0001, line 51: Can refactor to:
```javascript
profileData.profilePic = profileData.profilePic ?? defaultProfilePoc;
```

see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -21,22 +21,49 @@ const ProfileInfoEdit = () => {
   const dispatch = useDispatch();
 
   const {
-    teamId, teamName, orgName, email,
+    teamId, teamName, orgName, email, profilePic,
```
aksh0001, line 24: You may want to define `profilePic` in the reducer as well: https://github.com/Researchify/Researchify/blob/237b45c4d74421cf45cb1517775fa305f5b9a0a9/client/src/reducers/teamReducer.js#L15-L24

unless you plan to integrate with Redux later?

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: Weird bug, doesn't display the defaultProfilePic if I use this

client/src/components/profileInfoEdit/ProfileInfoEdit.js
rpra0006: Can it be refactored later? I need help from @pmyip3 but she has other tasks atm.

client/.eslintrc.json
```diff
@@ -1,5 +1,6 @@
 {
     "env": {
+        "browser": true,
```
jfan0010, line 3: duplicate of line 7

client/src/components/profileInfoEdit/ProfileInfoEdit.js
```diff
@@ -21,20 +21,46 @@ const ProfileInfoEdit = () => {
...
   } = useSelector(
     (state) => state.team,
   );
 
-  const [profileData, setInputs] = useState({ teamName, orgName, email });
+  const [setProfileData, setInputs] = useState({
```
jfan0010, line 29: I think what Akshay meant was changing `setInputs` to `setProfileData`, i.e.:
`const [ profileData, setProfileData] = useState ...`

jfan0010 (commented):
overall lgtm

client/src/components/profileInfoEdit/ProfileInfoEdit.js
aksh0001: yep just change of naming to maintain consistency

aksh0001 (approved):
LGTM on overall readability. Will defer to others on the actual functionality.

Also, for displaying the image bytestring, this may help: https://stackoverflow.com/questions/42395034/how-to-display-binary-data-as-image-in-react

But IIRC in the memories application tute I sent you, we didn't have to do anything special: https://youtu.be/ngc9gnGgUdA?t=3935. They showed the image in a `CardMedia` component from Material.

jfan0010 (approved):
LGTM. Can add `style={{ objectFit: 'cover' }}` in `Image` or in css to keep the image as original ratio if you want to.

pmyip3 (approved):

PR #208, opened 2021-09-09
`Researchify/client` frontend refactor pt.1 

jfan0010:
# Description 
Enabled some eslint rules and refactor accordingly. Rules enabled in this PR:

- "camelcase": "off",
- "consistent-return": "off",
- "prefer-destructuring": "off",
- "no-shadow": "off",
- "no-await-in-loop": "off",

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
Eslint rules were suppressed before.

# Solution description
As above

# Tests 
SA Checks passed?
Yes
 
All unit tests pass?
N/A

Builds successfully? 
Yes

# Risk
Tiny

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -11,6 +11,7 @@
     "rules": {
         "max-len": "off",
         "no-underscore-dangle": "off",
+        "no-plusplus": "off",
         "no-use-before-define": ["error", {"functions":  false, "classes":  false}],
         "react/react-in-jsx-scope": "off",
         "react/jsx-filename-extension": "off",
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -12,7 +12,6 @@
         "max-len": "off",
         "no-underscore-dangle": "off",
         "consistent-return": "off",
-        "no-plusplus": "off",
         "prefer-destructuring": "off",
         "no-shadow": "off",
         "no-await-in-loop": "off",
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -11,7 +11,6 @@
     "rules": {
         "max-len": "off",
         "no-underscore-dangle": "off",
-        "camelcase": "off", 
         "consistent-return": "off",
         "no-plusplus": "off",
         "prefer-destructuring": "off",
@@ -32,6 +31,7 @@
         "no-param-reassign": "off", // should refactor later
         "no-return-assign": "off", // should refactor later
         "spaced-comment": "off",
-        "arrow-body-style": "off" // should refactor later
+        "arrow-body-style": "off", // should refactor later
+        "linebreak-style": "off"
     }
 }
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -11,7 +11,7 @@
     "rules": {
         "max-len": "off",
         "no-underscore-dangle": "off",
-        "no-plusplus": "off",
+        "no-plusplus": "off", // keep `var ++` for consistency
         "no-use-before-define": ["error", {"functions":  false, "classes":  false}],
         "react/react-in-jsx-scope": "off",
         "react/jsx-filename-extension": "off",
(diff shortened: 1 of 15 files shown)


client/src/components/publications/form/ImportSuccessPage.js
```diff
@@ -80,7 +80,7 @@ const ImportSuccessPage = ({ closeModal }) => {
     let checkedPublications = publications.filter(
       (pub, idx) => publicationsToImport[idx],
     );
-    for (let i = 0; i < checkedPublications.length; i++) {
+    for (let i = 0; i < checkedPublications.length; i += 1) {
```
aksh0001, line 83: Can we hold onto this rule? https://github.com/Researchify/Researchify/blob/main/api/.eslintrc.json#L16 we're doing the same for the API. For consistency and also I kinda prefer `++` rather than `+=`.

client/.eslintrc.json
```diff
@@ -32,6 +26,7 @@
         "no-param-reassign": "off", // should refactor later
         "no-return-assign": "off", // should refactor later
         "spaced-comment": "off",
-        "arrow-body-style": "off" // should refactor later
+        "arrow-body-style": "off", // should refactor later
+        "linebreak-style": "off"
```
rpra0006, line 30: https://github.com/Researchify/Researchify/pull/204#discussion_r703589285

rpra0006: Link to Akshay's comment on linebreak-style. Maybe someone with experience can answer this?

jfan0010: Hmm, cause if I don't turn it off, I'll get warnings in every file. But from limited research I did just now, the preferred way is to change the end of lines to `LF` to follow UNIX system. And we can change it by [changing the setting in vscode](https://qvault.io/clean-code/line-breaks-vs-code-lf-vs-crlf/). Anyways, I'll update this branch to change all the file to using `LF`

client/src/components/publications/form/ImportSuccessPage.js
jfan0010: Sure I'll turn this rule off

client/.eslintrc.json
```diff
@@ -2,7 +2,8 @@
     "env": {
         "commonjs": true,
         "es2021": true,
-        "node": true
+        "node": true,
+        "browser": true
```
jfan0010, line 6: In `client/src/components/landing-pages/landingPageComponents/landingPageLayouts/LandingHeader.jsx` (line 31 ++), variable `window` and `browser` are used, adding this line of code removes the warning message.

client/.eslintrc.json
vgoh1: See my comment I just posted: https://github.com/Researchify/Researchify/pull/204#discussion_r705197184
This should be ok.

client/.eslintrc.json
vgoh1: Thanks for this, I didn't know you could do it like this. I was just suppressing the rules line by line - could remove those comments as well?

client/.eslintrc.json
jfan0010: ok ill add it back

client/.eslintrc.json
jfan0010: Which comment are you refering to? If you're saying the comment in `line34`: 
```
33     const navbar = document.getElementById('navbar');
34     window.onscroll = function () { // eslint-disable-line func-names
35        const maxScroll = document.body.clientHeight - window.innerHeight;
36        const currentScrollPos = window.pageYOffset;
```
I think we still need to keep the comment to suppress unnamed function, unless we give the function a name?

vgoh1: if you do a search for `no-undef`, these come up:
image
for localstorage, window and document

jfan0010: okie I'll change them

aksh0001 (approved):

rpra0006 (approved):

vgoh1 (approved):

PR #209, opened 2021-09-09
`Researchify/client` eslint refactor pt.2 

jfan0010:
# Description 
Turn eslint "react/prop-types" back on. Props passing into a component need to be checked.
** This branch is being refactored base on [#PR208](https://github.com/Researchify/Researchify/pull/208), once that PR is merged, this PR might be smaller.

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
"react/prop-types" was suppressed before when introducing airbnb eslint.

# Solution description
As description

# Tests 
SA Checks pass? Yes

All unit tests pass? N/A

Builds successfully? Yes

Reffered to [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) when refactoring.

# Risk
Medium. Will raise warning in console if the `propType` is not identified correctly


client/src/components/achievements/Achievement.js
```diff
@@ -123,4 +124,9 @@ const Achievement = ({ achievement }) => {
   );
 };
 
+// props validation
+Achievement.propTypes = {
+  achievement: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
```
vgoh1, line 129: Would it have saved you time by just supressing `react/forbid-prop-types` in .eslintrc.json since it's used quite a few times?

vgoh1 (approved):

client/src/components/achievements/Achievement.js
jfan0010: Make sense... Will refactor them

aksh0001 (approved):
I'd assume we won't need `PropTypes` if we started off using TypeScript instead :rofl:

client/src/components/achievements/Achievement.js
```diff
@@ -123,4 +124,9 @@ const Achievement = ({ achievement }) => {
   );
 };
 
+// props validation
```
aksh0001, line 127: curious: is there a reason why we're leaving line comments above some of the `PropTypes` implementations and not on others? We could either leave comments on top of all of them or remove them entirely to maintain consistency.

jfan0010: The most possible reason is that I forgot to add on others. Will add them now XD

cyr17 (approved):

PR #216, opened 2021-09-12
Change the wording layout to groupby

pmyip3:
# Description 
As title 

# Type of change 
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
The word layout is a bit confusing 

# Solution description
Change the relevant files in api and client
** Haven't changed related files in scholly as the new scholly layouts haven't considered the saved sorting order from editor site, will fix this in a separate ticket. 

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

image
image

Change the group by option, it should work the same as the layout

# Risk
Low


scholly/base/src/shared/config/publications.js
```diff
@@ -9,19 +9,16 @@ const categoryType = {
   Book: 'Book',
   Other: 'Other',
 };
-const layoutOptions = {
-  ALL_PUBLICATION: 'All Publication',
-  BY_CATEGORY: 'By Category',
+const groupByOptions = {
+  NONE: 'None',
+  CATEGORY: 'Category',
```
vgoh1, line 14: Just wondering, do we need this in scholly/base if they're not being used? Or just for consistency? I think the only ones being used in the layouts are the page sizes.

vgoh1:
I think there might be an edge case for existing teams in this branch where you log in and it just shows `Group By:` without an option selected, but once I chose an option and saved it was fine. I also see that you renamed the existing docs in the DB so I'm not sure what could've caused it, but it doesn't break the site so it shouldn't be too much of an issue.

scholly/base/src/shared/config/publications.js
pmyip3: I was planning to change it to a separate ticket, but Scholly will need that for my next ticket to display the publication based on the sorting/ layout options.

pmyip3:
> s

It's because the options data haven't been returned from the API call. Previously, I was putting the default sorting/ layout in the redux state so it displays the default one instead of being blank, but I don't find it nice as the options will change automatically after the option data is returned from API call, so I just removed the default sorting/ layout in the redux state. If we are going to implement a loading page, then it should be fine no matter which way we use.

vgoh1 (approved):
Makes sense, approval from me.

aksh0001 (approved):

jfan0010 (approved):
Works on mine

PR #217, opened 2021-09-12
Fix deploy for accounts with no initial repositories

kdav108:
# Description 
Pretty much what the title suggests :)

Shout out to @melvinalex7 @rsal0007 @sanathmahesh for helping uncover this bug.

# Type of change 
- [X] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
Currently, the deploy functionality fails if the user has a GH account with no repositories or only private repositories (which doesn't include the GH pages repo: `.github.io`)

# Solution description
Check the error message thrown by GitHub when searching repositories, if it's `422 Unprocessable entity` it is because the user has no repositories.

# Tests 
All unit tests pass?
yes
Builds successfully?
yes

1) Try creating a fresh account and deploy through Researchify from `main` branch - it should fail, then try it on this branch and confirm it works

N/A

# Risk
low (existing stuff largely untouched)


scholly/service/core/push.js
```diff
@@ -96,7 +107,7 @@ async function makeGHRepo(ghUsername, ghToken, repoName) {
         headers: options,
         data: createRepoBody,
       });
-      console.log('Repo was created successfully');
+      winston.info('Repo was created successfully');
     } catch (err) {
       console.log(err);
```
vgoh1, line 112: Can we use winston here as well?

scholly/service/core/push.js
```diff
@@ -43,7 +43,7 @@ async function pushBuiltAppToPages(ghUsername, ghToken, repoName) {
       } else {
         winston.info(`Successfully deployed app for ${ghUsername}`);
       }
-      cleanupBuild(); // TODO this is not working for some reason
+      cleanupBuild();
```
aksh0001, line 46: Does this work now?

scholly/service/core/push.js
```diff
@@ -68,22 +68,33 @@ async function makeGHRepo(ghUsername, ghToken, repoName) {
...
       },
     );
+    repoExists = repoResponse.data.total_count === 1;
   } catch (err) {
-    console.log(err);
+    if (err.response.status !== 422) {
+      // Error 422 is fine (it can be caused by user having no repo or only private ones in their account), any other should be thrown back up.
+      winston.error(
+        `Error checking existing repositories for user: ${ghUsername}`,
+        err,
```
aksh0001, line 86: I believe passing `err` as an arg to the logger doesn't do what you're trying to do here. We might have to interpolate the `err.message` as part of the initial logging message.

scholly/service/core/push.js
```diff
@@ -68,22 +68,33 @@ async function makeGHRepo(ghUsername, ghToken, repoName) {
...
+      );
+      throw err;
```
aksh0001, line 88: nit: avoid log and re-throw, https://softwareengineering.stackexchange.com/questions/365427/try-catch-log-rethrow-is-anti-pattern

scholly/service/core/push.js
```diff
@@ -68,22 +68,33 @@ async function makeGHRepo(ghUsername, ghToken, repoName) {
...
       {
         headers: options,
       },
     );
+    repoExists = repoResponse.data.total_count === 1;
   } catch (err) {
-    console.log(err);
+    if (err.response.status !== 422) {
+      // Error 422 is fine (it can be caused by user having no repo or only private ones in their account), any other should be thrown back up.
+      winston.error(
```
aksh0001, line 84: also, nit: can we import the name as `logger` rather than `winston` up the top of this file for consistency?
so that log statements would become `logger.error`, etc.

aksh0001 (commented):
Also, this may be for another PR, but I'd prefer if we wrapped requests to GitHub with a higher-level library rather than issuing raw requests. Have you considered libraries like: https://www.npmjs.com/package/github-api, https://github.com/octokit/rest.js/, etc. ?

kdav108:
> Also, this may be for another PR, but I'd prefer if we wrapped requests to GitHub with a higher-level library rather than issuing raw requests. Have you considered libraries like: https://www.npmjs.com/package/github-api, https://github.com/octokit/rest.js/, etc. ?

Sure, not against it but would like to know what's the benefit in using those libraries? The raw request don't seem to have complexities or add much repeated code, so don't see any direct benefit from using a library instead.

scholly/service/core/push.js
kdav108: Interesting discussion. In fact the post doesn't seem to have much unanimous conclusion. My personal opinion (and I'm no design expert) is that adding a debug message there makes life easier. I have removed this though, because in this particular case the message isn't as useful.

kdav108: Actually to think of it again, if I want to show a message I should actually raise a custom Error (possible preserving some information from the caught error), that way whoever calls this function will get the message as well (and a dev won't have to log through the logs to find it). So I guess I agree 😅

scholly/service/core/push.js
aksh0001: SGTM, resolving

scholly/service/core/push.js
aksh0001: Yeah, I believe the idea is to avoid unnecessary log entries. Maybe, you could add more context to this `err` by inserting a message? That might get you what you want - encapsulate extra info about the error while following this suggestion.

scholly/service/core/push.js
```diff
@@ -146,13 +152,13 @@ async function createPagesSite(ghUsername, ghToken, repoName) {
       data: pagesBody,
     });
     if (createPagesResponse.status === 201) {
-      console.log('Pages site successfully created');
+      logger.log('Pages site successfully created');
```
aksh0001, line 155: nit: do you mean `logger.info`?

scholly/service/core/push.js
```diff
@@ -146,13 +152,13 @@ async function createPagesSite(ghUsername, ghToken, repoName) {
...
     } else {
       // see https://docs.github.com/en/rest/reference/repos#create-a-github-pages-site
-      console.log(createPagesResponse.data);
+      logger.log(createPagesResponse.data);
     }
   } catch (err) {
-    console.log(err);
+    logger.log(err);
```
aksh0001, line 161: logger.error?

scholly/service/core/push.js
```diff
@@ -172,9 +178,9 @@ async function buildPages(ghUsername, ghToken, repoName) {
       method: 'post',
       headers: options,
     });
-    console.log('Build queued');
+    logger.info('Build queued');
   } catch (err) {
-    console.log(err);
+    logger.info(err);
```
aksh0001, line 183: .error()?

scholly/service/core/push.js
```diff
@@ -146,13 +152,13 @@ async function createPagesSite(ghUsername, ghToken, repoName) {
       data: pagesBody,
     });
     if (createPagesResponse.status === 201) {
-      console.log('Pages site successfully created');
+      logger.log('Pages site successfully created');
     } else {
       // see https://docs.github.com/en/rest/reference/repos#create-a-github-pages-site
-      console.log(createPagesResponse.data);
+      logger.log(createPagesResponse.data);
```
aksh0001, line 158: Should we make this a debug level log, if printing out the response isn't strictly necessary?

aksh0001 (approved):
LGTM

aksh0001:
> > Also, this may be for another PR, but I'd prefer if we wrapped requests to GitHub with a higher-level library rather than issuing raw requests. Have you considered libraries like: https://www.npmjs.com/package/github-api, https://github.com/octokit/rest.js/, etc. ?
> 
> Sure, not against it but would like to know what's the benefit in using those libraries? The raw request don't seem to have complexities or add much repeated code, so don't see any direct benefit from using a library instead.

Yeah, so I think the existing code is fine and as you said it isn't that complex. The reasons why I'd opt for a higher level library in this case:
- It further abstracts away implementation detail - e.g. what API endpoint to call, request parameters to use, structure of body, etc.
- The API abstractions are more meaningful than issuing raw axios requests. E.g. see the one for creating a commit: https://octokit.github.io/rest.js/v18#git-create-commit, the methods names describe intention very clearly
-  More importantly, using a library means the onus is on the vendor of the library to roll out any updates and patches if GitHub's API changes, or if something else is wrong with their endpoints. By issuing requests ourselves, we are coupled with a specific implementation using pre-configured endpoints, that may change in the future (granted, that's not very likely, but having them maintain the package is a plus)

rpra0006 (approved):
Looks good

melvinalex7 (approved):
Tested Deploying the Client Website into GitHub account with no repositories. Looks like it works fine. Great work.

scholly/service/core/push.js
kdav108: hmmm interesting, I vividly remember fixing this. but anyways, thanks for catching it, I'll fix it now

PR #229, opened 2021-09-19
Delete batch publications

pmyip3:
# Description 
1. New feature to allow users to delete more than 1 publication at once.  
2. After import publications, revert the header color of all the imported publications to the original color at the same time. 

# Type of change 
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 
- [x] Improvemnt  

# Problem
1. User needs to delete publication once at a time
2. After import publications, only the publications on the current page revert back to the original color after 2.5 seconds, the  publications that aren't on the current page still have the props of 'newlyAdded', which cause the header to remain green when the user clicks on that page  

# Solution description
1. Add deleteBulkPublications endpoint to handle deleting more than one publication at a time. 
2. Dispatching an action which removes the 'newlyAdded' props for all publications after 2.5 seconds 

# Tests 
All unit tests pass? N/A

Builds successfully? Yes

image
image

# Risk
Low


vgoh1:
Does this happen on your end? After creating a new publication, the new publication doesn't show up in the publications list anymore despite the number of publications changing. It shows up after I refresh the page though and after I import publications.

cyr17 (approved):
looks good!

vgoh1 (approved):
The issue I was having is gone, so LGTM.

rpra0006 (changes requested):
Delete publications button doesn't work when I click on it
image

pmyip3:
> Delete 10 publications button doesn't work when I click on it. Single publication delete works though.
> image

No idea why the onClick function was removed, but it should be fixed now, thanks for testing it.

rpra0006 (approved):
All working now

api/src/controllers/publications.js
```diff
@@ -314,11 +314,30 @@ function importPublications(req, res, next) {
...
 
+/**
+ * Handles a PATCH request to delete a list of publication by the mongo object id on the endpoint /publications/
+ *
+ * @param req request object - the list of publication ids given in the body
+ * @param res response object
+ * @returns 200: publications deleted successfully
+ * @returns 400: error deleting publications
+ */
+async function deleteBulkPublications(req, res, next) {
```
aksh0001, line 325: nit with function name. I think `batch` delete is more common than `bulk` delete in terms of naming this kind of functionality (as with the PR name too). Also, curious about why `PATCH` and why not `DELETE`?

client/src/actions/publications.js
```diff
@@ -10,6 +10,7 @@ import {
   IMPORT_FAIL,
   IMPORT_END,
   IMPORT_EMPTY,
+  DELETE_BULK_PUBLICATIONS,
```
aksh0001, line 13: Same naming here if we go with -> `BATCH`

aksh0001: ditto everywhere else

client/src/actions/publications.js
```diff
@@ -151,3 +153,16 @@ export const createBulkPublications = (teamId, publicationList) => async (dispat
...
 };
+
+export const deleteBulkPublications = (publicationIdList) => async (dispatch) => {
+  try {
+    await api.deleteBulkPublications(publicationIdList);
+    dispatch({
+      type: DELETE_BULK_PUBLICATIONS,
+      payload: publicationIdList,
+    });
+    dispatch(successMessageCreator(`${publicationIdList.length} publication(s) has been deleted`));
```
aksh0001, line 164: nit `has` -> `have`

client/src/components/publications/Publications.js
```diff
@@ -98,6 +99,15 @@ const Publications = () => {
   useEffect(() => {
     const sortedPublication = sortPublications(teamPublications, options.sortBy);
     setPublications(sortedPublication);
+    const newlyAddedPublications = teamPublications.filter((pub) => pub.newlyAdded === true);
```
aksh0001, line 102: do we need to `=== true`? or can we omit it and just have `(pub) => pub.newlyAdded`? Also, this is a separate issue, but newlyAdded could be renamed to `isNewlyAdded` to make it more clear what we're referring to.

aksh0001 (approved):
LGTM, only minors/nits

api/src/controllers/publications.js
pmyip3: cause the delete api call only accept a parameter (the deleted document id), but do not accept body

PR #244, opened 2021-09-24
Omit eslint config in Dockerfiles by placing in dockerignore

aksh0001:
# Description 
Building the scholly base app in prod was causing issues as the `airbnb` eslint configuration was not ported over there due to running in `NODE_ENV=production` (airbnb styles are a dev dependency). This change will ignore those files when building the images so that they don't exist in the container filesystem.

# Type of change 
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
See description.

### DIFF
--- a/api/.dockerignore
+++ b/api/.dockerignore
@@ -1,1 +1,2 @@
 node_modules
+.eslintrc.json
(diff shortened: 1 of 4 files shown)


kdav108 (approved):
Thanks for the fix!

PR #254, opened 2021-10-01
Remove eslint-config-airbnb from eslintConfig.extends in client

vgoh1:
# Description 
This is a fix for the issue where we are failing to build the Docker image for the client due to eslint throwing errors. See: https://github.com/Researchify/Researchify/runs/3763005968
This only happens when building the image and not locally for some reason, perhaps it's not copying over the `.eslintrc.json` or it can't find it, which is supposed to suppress the rules.

# Type of change 
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
See description

# Solution description
See https://github.com/Researchify/Researchify/pull/254#issuecomment-932936543

# Tests 
Build the docker image in `/client` by running `docker build .` and it shouldn't throw eslint errors.

N/A

# Risk
Low


vgoh1:
@aksh0001 @kdav108 Just curious, do we know if this is an issue for scholly/base when deploying to GH pages? Does it need to be added there as well?

aksh0001:
> @aksh0001 @kdav108 Just curious, do we know if this is an issue for scholly/base when deploying to GH pages? Does it need to be added there as well?

We're ignoring the eslint config for `scholly/base` via `.dockerignore`, so it shouldn't be copied over to the container and the building should be successful still if i'm not wrong. I'd like to look into this further though before merging because this was all working fine without any hacks a couple weeks ago :disappointed: 

Also was wondering if we installed eslint correctly in the client and base? 
https://github.com/airbnb/javascript/issues/2233#issuecomment-754316789

the docs say to use `npx install-peerdeps --dev eslint-config-airbnb` https://www.npmjs.com/package/eslint-config-airbnb

pmyip3 (approved):

pmyip3 (approved):

jfan0010 (approved):

vgoh1:
@aksh0001 So I did a bit more investigation and found out that we don't get the same behaviour in scholly/base because in that `package.json`, eslint-config-airbnb was not included in the[ eslintConfig.extends array](https://github.com/Researchify/Researchify/blob/main/scholly/base/package.json#L32). However, it is in the [client package.json](https://github.com/Researchify/Researchify/blob/f5b822d6a1cde5d9fe645deb2862b3d8998bd02e/client/package.json#L48). I don't remember why I didn't include it in scholly/base at the time though, may have been reading a different tutorial that told me to set it up differently ... Additionally, after reading the [eslint docs](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats), `.eslintrc.json` takes precedence over what's in the `package.json`. 

This would line up with how the deployments started breaking after the `eslintrc.json` files got ignored by Docker in [main](https://github.com/Researchify/Researchify/runs/3707854895), because the rules its complaining should just be regular eslint complaining - which leaves the question why it's not complaining about how it can't find the airbnb config like from when scholly/base wouldn't build because it couldn't find the airbnb config to extend from, as it was a dev dependency.

I think the fix should be removing `eslint-config-airbnb` from the eslintConfig.extends array in `package.json` but I'm running  into another error when doing `npm run build` which I'm looking into at the moment.

client/package.json
```diff
@@ -22,7 +22,7 @@
     "react-dom": "^17.0.2",
     "react-github-login": "^1.0.3",
     "react-hot-toast": "^1.0.2",
-    "react-icons": "^4.2.0",
+    "react-icons": "4.2.0",
```
vgoh1, line 25: react-icons v4.3.0 is broken and was causing npm run build and start to fail, and the fix is to force the older version, see https://github.com/react-icons/react-icons/issues/490#issuecomment-932823191 
also checked that the package-lock.json is using v4.2.0

aksh0001: Nice catch. Luckily it seems they just fixed it yesterday with 4.3.1:
https://github.com/react-icons/react-icons/issues/490#issuecomment-932996600

So I think we can just leave it at ^4.2.0.

aksh0001 (approved):

PR #260, opened 2021-10-04
Backend Team Delete

aksh0001:
# Description 

We have middleware for triggering dependent document creation for team registrations. This PR implements a dependent document deletion middleware/hook/trigger to delete associated documents when a team has been deleted.

This PR is a result of comments left in https://github.com/Researchify/Researchify/pull/198.

Note: I've also changed up the positioning of the controller and routes to follow CRUD, so we define create endpoints first, followed by read, update, and then delete (the additions are few, but the diff count is bloated because of this) to make it easier to follow. The only additions are the `deleteTeam()` controller + route, and changes in the `Team` model.

# Type of change 
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation 

# Problem
There is no way to delete a team in the backend.

# Solution description
Implement a delete method for the backend, as well as a post hook that will delete any associated documents.

# Tests 
All unit tests pass? Y

Builds successfully? Y

Screenshot from 2021-10-04 00-07-53

Returns `204` (no content)

Please manually verify that the dependent documents have been deleted in the database.

# Risk
Low, additive.

### DIFF
--- a/api/src/controllers/team.js
+++ b/api/src/controllers/team.js
@@ -16,9 +16,186 @@ const {
 const { fillErrorObject } = require('../middleware/error');
 
 /**
+ * Handles a POST request to create a team on the endpoint /team.
+ * A post-hook/middleware is associated with the {@link Team} model that will
+ * create default-initialized dependent documents for the team being created.
+ *
+ * @param req request object containing at least two fields: teamName & orgName.
+ * @param res response object - updated team object
+ * @param next handler to the next middleware
+ * @returns 201: returns updated team details
+ */
+async function createTeam(req, res, next) {
+  const foundTeam = await Team.findOne({ email: req.body.email });
+  if (foundTeam) {
+    return next(
+      fillErrorObject(400, 'Duplicate email error', [
+        'Email had been registered',
+      ]),
+    );
+  }
+  const salt = await bcrypt.genSalt();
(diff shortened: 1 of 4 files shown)


api/src/controllers/team.js
```diff
@@ -183,6 +172,78 @@ function updateTeamMember(req, res, next) {
...
+    // validate the handle by getting user id
+    if (!process.env.TWITTER_BEARER_TOKEN) {
+      return next(
+        fillErrorObject(500, 'Missing environment variable', [
+          'No Twitter API Bearer Token found in .env file',
+        ]),
+      );
+    }
+    const response = await axios.get(
+      `https://api.twitter.com/2/users/by/username/${handle}`,
```
cyr17, line 222: Out of curiosity, Is octokit just for raw gitHub requests or can it be used here too?

aksh0001: Octokit is the official GitHub client library. But good spot, we should upgrade this code to use the official Twitter library in the future.

I haven’t used it before in NodeJS but this seems to be the one https://www.npmjs.com/package/twitter-api-v2

jfan0010 (approved):
lgtm

cyr17 (approved):
Allg , might need to fix up eslint

pmyip3 (approved):

rpra0006 (approved):

PR #235, opened 2021-09-21
Logout Info Dropdown on Dashboard

cyr17:
Co-Authored-By: @melvinalex77 
# Description 
A summary of the change 
Added logout, settings as part of a dropdown accessed by clicking the profile button on the top left part of the header
Styled the button and the Researchify logo to match the landing page

# Type of change 
- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation 

# Problem
Briefly describe the root cause and analysis of the problem 
The current sidebar is cluttered, and the account buttons position is hard to track so made it fixed to the bottom and moved the logout to the dropdown button

# Solution description
Describe your code changes in detail for reviewers. 
Explain the technical solution you have provided and how it fixes the issue.

# Tests 
All unit tests pass?
yes
Builds successfully?
yes

Attach screenshot of the expected result (e.g. UI of the website/ postman api call result/ database changes)
image

Provide steps in detail for reviewers to reproduce the expected result of the code changes.
Click on the drop-down xp.

# Relevant document/ link (if any) 
Any related documents could help reviewers to review this PR 
Refer to these trello tickets
https://trello.com/c/faLXmHnM
https://trello.com/c/PYQJNf1g

# Risk
Identity the risk level (low, medium, high) and indicate which component(s) will be affected if bugs exist.
low

PR #261, opened 2021-10-04
narrow view twitter

jfan0010:
# Description
Implement a floating Twitter box to display Twitter in a narrow view.

# Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Improvement

# Solution description
Add a Twitter icon at the bottom right as a button to trigger the Twitter box. `z-index` to float.
 
# Tests
All unit tests pass? No unit test anyway
Builds successfully? y
Do all SA checks pass? y

In a narrow screen, twitter icon will shown when twitter box disappear due to screen size:
narrow1

When clicked: 
float twitter box

# Risk
Moderate. might affect existed UI component


aksh0001:
Just noticed some console errors, but not entirely sure if its from this PR?

Screenshot from 2021-10-04 13-49-47

jfan0010:
> Just noticed some console errors, but not entirely sure if its from this PR? Screenshot from 2021-10-04 13-50-39

I think it's from [PR#235](https://github.com/Researchify/Researchify/pull/235), will look into it and try fix in another PR.

aksh0001 (approved):
maybe it's do with my screen, but is anyone else's `Tweets` title section squished in with the top bar margin?

jfan0010:
> maybe it's do with my screen, but is anyone else's `Tweets` title section squished in with the top bar margin? 

I think that is because in current version, when we click the button to open twitter box, we use the current window height to decide the height of the twitter box. And when you're changing window size while it is opened, it won't change dynamically, but close and open again will change the height. 

If I was correct, the user shouldn't have a problem with this when they are not changing the screensize while using researchify. Anyways I have another ticket to improve the height changing mechanism, which hopefully will fix this problem.

rpra0006 (approved):
Runs smoothly on my end. Just my opinion though, would the twitter button be more accessible if it was light blue? Maybe we can stick with the same color for border, but a light blue inside for the bird.
image
image

jfan0010:
> Runs smoothly on my end. Just my opinion though, would the twitter button be more accessible if it was light blue? Maybe we can stick with the same color for border, but a light blue inside for the bird. 

Yes I was thinking this as well when I'm implementing, but then I go for consistency instead of accessibility, what does everyone think?

vgoh1 (approved):
I like how the Twitter bird fits the color scheme! I dunno about how making the bird light blue would be better for accessibility though.

PR #268, opened 2021-10-05
Move email and password to top of register form

pmyip3:
# Description
As title 

# Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Improvement

# Solution description
Sign up form should have the most important content at the top (email, password)

# Tests
All unit tests pass? N/A
Builds successfully? Yes 
Do all SA checks pass? Yes

image

N/A

# Risk
Low

### DIFF
--- a/client/src/components/auth/authComponents/RegistrationForm.jsx
+++ b/client/src/components/auth/authComponents/RegistrationForm.jsx
@@ -97,36 +97,6 @@ const RegistrationForm = () => {
           handleSubmit, handleChange, values, touched, errors,
         }) => (
           <Form noValidate onSubmit={handleSubmit}>
-            <Form.Row>
-              <Form.Group as={Col} md="6">
-                <Form.Label> Team name </Form.Label>
-                <Form.Control
-                  type="text"
-                  name="teamName"
-                  placeholder="Team name"
-                  value={values.teamName}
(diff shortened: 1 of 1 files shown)


aksh0001 (approved):
LGTM. 

Side note: were we planning to get rid of the org name based on UAT as well, iirc someone said they were confused between org name vs. team name?

rsal0007 (approved):

jfan0010 (approved):
Personally prefer leaving the name on top, and I think that's what most websites do (like Gmail and Facebook). But lgtm if that's what users prefer to.
gmail
facebook

vgoh1:
@jfan0010 I feel like it depends on the type of application, gmail and facebook have name first because the application is reliant on the user's name. You can argue team name and org name is important for us though. What do the others think? 
In the future we could have setting team name and org name as part of onboarding after they sign up? Cause that would deal with the confusion between team and org name.
@aksh0001 I feel like there's still some discussion needed on whether to remove org name or not, because research groups tend to be tied to an uni but our sign up form isn't clear on what it means by org. And I think it's also because we go against convention mentioned by Jie above by asking for team name rather than the user's name that adds more confusion.

PR #269, opened 2021-10-06
Improvements to page selector + disabled button styling

vgoh1:
# Description
A suggestion we got from UAT was to add descriptions to the page selector so that users know what each page is for. Also had the idea to let the user to select multiple pages in one go instead of multiple times to make it more streamlined. Since it was a small change, also added styling for a disabled button since that was lost after we started using styled components.

# Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Improvement

# Solution description
Used cards to show multiple pages in the page selector
Managing CSS is easy with styled components
Also tweaked add page endpoint a bit to handle an array instead of just a string

# Tests
Disabled button:
image

Page selector:
image

Disabled button tooltip:
image

N/A

# Risk
Low, just page selector is affected


api/src/controllers/website.js
```diff
@@ -40,7 +40,7 @@ async function addWebPage(req, res, next) {
   try {
     const foundWebsiteInfo = await Website.findOne({ teamId });
     if (foundWebsiteInfo) {
-      foundWebsiteInfo.pages.push(req.body.pageName);
+      foundWebsiteInfo.pages = foundWebsiteInfo.pages.concat(req.body.pageName);
```
aksh0001, line 43: does the former not work?

aksh0001 (commented):
just my 2c on the UI: instead of having checkboxes on the cards, could we make the cards clickable themselves, so that when selected, the card is emphasized (E.g. highlighted, margin emboldened,  color changed, etc.) - i've seen this type of interface quite often, but if it's difficult to implement, this should be ok

api/src/controllers/website.js
vgoh1: I should've renamed `pageName` to `pageArray` so that it's less confusing, but the reason why push doesn't work is because `pageName` became an array instead of just the page name. So it would be pushing an array into an array rather than a string into the array. I'll rename the variables accordingly.

pmyip3 (approved):
LGTM

jfan0010 (approved):

PR #277, opened 2021-10-11
Publication page improvement

pmyip3:
# Description
Make publication page consistent with team and achievement page 

# Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [x] Improvement

# Solution description
- move the delete batch button next to add publication button
- make css styling consistent with team and achievement page 

# Tests
All unit tests pass? N/A
Builds successfully? Yes
Do all SA checks pass? Yes

image

N/A

# Risk
Low, UI notification only


client/src/components/publications/publication/Publication.js
```diff
@@ -160,9 +167,9 @@ const Publication = ({ pub }) => {
             {
               isHovering
             && (
-            <StyledButtonGroup className="float-right">
+            <StyledButtonGroup onClick={childCallback} className="float-right">
```
aksh0001, line 170: curious: do we actually need an `onClick` here? If the whole purpose of this is to prevent the click event from   being propagated up the DOM, why not just not have the `onClick` event?

rsal0007 (approved):

client/src/components/publications/publication/Publication.js
pmyip3: Cause I have the onClick event on its parent div (line 147), and I don't want to trigger that event when its children div are clicked (line 171 and 172) cause the children div has their own onClick event. 
Not sure if it is the correct way to do it but I found this solution from here. 
https://stackoverflow.com/questions/37568550/react-prevent-event-trigger-on-parent-from-child

ssha0054 (approved):
Looks good! I like the link icon being at the start as it is right now, feel like if it's at the end then it won't be as nice since most publication titles will be of different length so the positioning of the icon won't be consistent.

PR #282, opened 2021-10-13
Favicon For Client Websites + Researchify

melvinalex7:
# Description

A summary of the change
Researchify Logo is now set as the client website's favicon
# Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Improvement

# Solution description

# Tests

All unit tests pass?
Y
Builds successfully?
Y
Do all SA checks pass?

image

- To test in development environment, use npm run dev in scholly/base. The favicon for all layouts will be the Researchify Logo.

# Risk


sanathmahesh (approved):
Works and Looks good, might need to make it so the user can change the logo since it's the deployed site, but it's a step up from the react logo.

melvinalex7:
> Works and Looks good, might need to make it so the user can change the logo since it's the deployed site, but it's a step up from the react logo.

Yea even I thought about that, but the issue is right now the Researchify website does not explicitly say to upload logo. So the user may upload their logo or a group photo team. So it may not look good if it becomes a favicon.

ssha0054 (approved):
Looks good! Agree that it would be nice if the users can use their own logo but also agree that it is outside the scope of this PR. Having the Researchify logo looks better than the default React logo we had before.

aksh0001 (approved):
Curious why we're not just replacing the default favicon? https://medium.com/@jenniferdobak/the-public-folder-and-favicons-in-create-react-app-8dc2cc1d492b

melvinalex7:
> > Yea even I thought about that, but the issue is right now the Researchify website does not explicitly say to upload logo. So the user may upload their logo or a group photo team. So it may not look good if it becomes a favicon.
> 
> How about we add a place in Researchify asking users to add their own logo? Alternatively, we can have favicons for all the alphabets and we choose the corresponding favicon that matches the first letter of the team name.
> 
> And if we decided to only use Researchify 'R' for the client site favicon, is there a reason we don't just replace the favicon.ico in base/public folder?
> 
> This might be extra, but is it possible that we can change the icon to a more clear icon? not sure if it was just me but this 'R' icon looks a bit blur on my screen.

-> Well that's definitely can be a great addition in the future given that the logo uploaded is restricted to a square image and we can use a third-party favicon creator to convert the image into an icon.

-> Well, there is no particular reason that I chose the react-favicon but it definitely becomes easier to modify when the client websites may need to dynamically change logos based on the research group logos. I've replaced the images favicons in the public folders so that our logo will be the default at all times. 

-> I've increased the quality of the favicon, you can check it out.

PR #285, opened 2021-10-14
Integration tests on achievements controllers

ssha0054:
# Description

Note: This PR is low priority and does not need to be merged before the demo, so if you’re limited with time then I’d recommend reviewing another PR with higher priority instead of this for now :)

This PR adds integration tests on the achievements routes. 

# Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Improvement

# Solution description

Added tests using `jest` and `supertest` for the API endpoints on the achievements routes. More detailed information regarding the test setup can be found on the [readme](https://github.com/Researchify/Researchify/tree/main/api/test#readme) in the API/Test directory.

# Tests

All unit tests pass? Yes

Builds successfully? Yes

Do all SA checks pass? Yes

You will need to have mongodb installed locally to run the tests. A link in the [readme](https://github.com/Researchify/Researchify/tree/main/api/test#readme) is provided with a guide on how to install mongo locally.
Once you have mongodb installed, start it up and then do `npm run test` in the `api` directory. After the tests are done, stop running mongodb locally.

# Risk

Low, does not effect functionality.


aksh0001 (approved):
LGTM.
I think one downside to this approach of cookie/token passing is that when the tokens (refresh/access) or cookies storing these tokens expire, the test cases may likely fail. But, the chances of this happening under our current setup are quite rare (since access token expiry is 15min and refresh token expiry is 3d). Nevertheless, these can be changes very easily and cause failures. I've been doing some research and think a more robust solution would be to hold onto a non-expiring JWT specifically for tests to make it more deterministic. But we don't have time to investigate this, and this approach mimics real user interaction, so LGTM!

PR #286, opened 2021-10-14
Convert deploy into a floating button/Modal

kdav108:
# Description

With the current design of the deploy flow, we had 2 issues:
1) Needing to move back to the Dashboard page to deploy was a pain point for users
2) No information provided on how we deploy and what the deployed sit url is

This PR fixes these issues by creating a floating button for deploy that is available on all pages (that affect the deployed site). Upon clicking the user sees a Modal with more information. I have also combined the `Login with gitHub` and `Deploy with GitHub` button into just one, reducing the number of clicks required.

Also included in this a minor fix:
- When deploying site with no aboutUs description, the deployed code will fail, I've made it so that the attribute is defaulted to an empty string.

# Type of change

- [ ] Bug fix
- [X] New feature
- [X] Refactoring
- [ ] Documentation
- [ ] Improvement

# Tests

All unit tests pass?
Yes
Builds successfully?
Yes
Do all SA checks pass?
Yes
Deploy Modal:
Screen Shot 2021-10-14 at 9 52 21 pm

Floating button:
Screen Shot 2021-10-14 at 9 52 40 pm

1) Login and try deploying the site from different pages.
2) Try using the url shown on the Deploy Modal.

# Risk

medium, may impact deployment functionality


cyr17 (approved):
Looks Good

rpra0006 (approved):
Front-end functionality deploy button looks good. I'm just getting an error 500 when deploying, not sure if it's related to this PR or not.

client/src/components/dashboard/deploy/Deploy.js
```diff
@@ -0,0 +1,162 @@
...
+//     border: 1px solid #AB9671 !important;
+//     border-radius: .25rem;
+//     background-color: #AB9671;
+//     color: white;
+//     transition: color 0.15s ease-in-out, background-color 0.15s ease-in-out, border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
+//     font-size: ${(props) => props.fontSize}; //.875rem;
+//   &:hover{
+//     background-color: rgb(119, 105, 79);
+//   }
+// `;
```
aksh0001, line 38: nit: remove?

aksh0001: also, this may be part of the shared components as well

client/src/components/dashboard/deploy/Deploy.js
```diff
@@ -0,0 +1,162 @@
+import React, { useState } from 'react';
```
aksh0001, line 1: Could you add file docs here?

aksh0001 (dismissed):

vgoh1:
I love the fact that deploying is more streamlined! But I think there's a few places where the logic can be improved:
1. It looks like you're doing the access token exchange every time the user clicks on deploy - that is, we get a new access token each time and you can see the new tab that takes you to the GH login page. We should be caching/storing the initial access token on the frontend (like we were doing in localStorage) because:

- It would save us an extra call to GH API for the access token every time we want to deploy (also the pop up is annoying)
- We're lucky that GH doesn't have a TPS limit for the exchange flow, but it's good practice to cache anyway for situations when you encounter something that does have a TPS limit

2. Once the above is done, we can think about only showing the deploy modal the first time (which will prompt the login) and replace the hovering button with the spinner? This can be a later improvement though.

vgoh1:
Also, the deploy button is unclickable on the homepage editor, possibly because it overlaps with the Twitter box.

aksh0001:
Just found a bug that occurs if we're no longer storing the access token in local storage. Looking into this in the context of #198

client/src/components/dashboard/Dashboard.js
```diff
@@ -44,40 +41,33 @@ const Dashboard = () => {
...
+                availablePages={availablePages}
+                loading={loading}
+              />
+            </Tab>
+            <Tab eventKey="theme" title="Theme" className="pb-5">
+              <TemplateSelector teamId={teamId} />
+            </Tab>
+          </Tabs>
+        </Container>
+      </main>
```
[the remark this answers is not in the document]
kdav108: I've remove the tab overall, no need for it anymore

kdav108:
Thought I can get this in by Sunday before our submission but the issue of the deploy button not clickable took quite some time (the answer was `zIndex`).

Anyways, if you're still interested, @vgoh1 I've fixed the issues you identified

vgoh1:
the power of zindex :P

looks good to me, thanks for making those changes!

vgoh1 (approved):

api/src/controllers/team.js
```diff
@@ -430,7 +430,20 @@ async function deployToGHPages(req, res, next) {
   }
 
   const ghUsername = data.login;
+  const webUrl = `${ghUsername}.github.io`;
+  console.log(webUrl);
```
vgoh1: remember to remove this console log and others

client/src/components/dashboard/deploy/Deploy.js
```diff
@@ -0,0 +1,180 @@
...
+  position: { },
+};
+
+function mapStateToProps(state) {
+  return {
+    webUrl: state.url,
+  };
+}
+
+export default connect(mapStateToProps)(DeployBtn);
```
aksh0001: use hooks?https://www.samdawson.dev/article/react-redux-use-selector-vs-connect

api/src/controllers/team.js
```diff
@@ -430,7 +430,20 @@ async function deployToGHPages(req, res, next) {
   }
 
   const ghUsername = data.login;
+  const webUrl = `${ghUsername}.github.io`;
```
aksh0001, line 467: Ideally the web url should be returned by scholly service to avoid duplication of urls, but this is fine

aksh0001 (approved):

api/src/controllers/team.js
kdav108: ahhh there's always one I forget

client/src/components/dashboard/deploy/Deploy.js
kdav108: I was using useSelector before but it was hard to make the component remount when the redux state changes. There were then two solutions either to use connect() or using useEffect to link useSelector and a useState hook (https://stackoverflow.com/questions/59172453/should-i-pass-useselector-to-usestate), just went with connect to learn about it. 

What do you think? Worth changing it?

api/src/controllers/team.js
kdav108: hmmm I was thinking Scholly service should get the url from api, this makes it easier for us to extend functionality later to allow the user to customise their url.

aksh0001: makes sense

client/src/components/dashboard/deploy/Deploy.js
aksh0001: That's interesting. AFAIK `mapStateToProps()` and `useSelecter()` should behave identically.

client/src/components/dashboard/deploy/Deploy.js
```diff
@@ -0,0 +1,180 @@
...
+  webUrl: PropTypes.string.isRequired,
+  position: PropTypes.object,
+};
+DeployBtn.defaultProps = {
+  position: { },
+};
+
+function mapStateToProps(state) {
+  return {
+    webUrl: state.url,
```
aksh0001: Could be wrong, but should this not be `state.website.url`? Does it work as it is? Could also solve re-rendering issue with hooks

kdav108: ahhh yeah not sure how I missed that, it was a mistake during implementing the connect, I had it as `state.website.url` when I was trying with hooks :)

client/src/components/dashboard/deploy/Deploy.js
kdav108: `mapStateToProps()` sends the redux state as a prop to our component, so when the redux state changes the prop changes causing a re-render. With `useSelector` by itself it doesn't cause a re-render. We would have to use `useEffect` that depends on the variable set to `useSelector(state => state.website.url)`. So possible, just though I'd try `connect()`

aksh0001: https://react-redux.js.org/api/hooks#useselector

Gotcha, SGTM. I think `useSelector` is used for re-renders, if state changes.

see docs:

Screenshot from 2021-10-31 10-03-46

Screenshot from 2021-10-31 10-00-42

PR #290, opened 2021-10-17
Onboarding for Publications page

sanathmahesh:
# Description

Added onboarding for the publications page, the 2 forms for adding publications(import and manual), and for individual publication cards. Not automatic, need to be accessed using the help icons

# Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Improvement

# Solution description

# Tests

All unit tests pass?
Yes
Builds successfully?
Yes
Do all SA checks pass?
Yes

Screenshot (67)

Screenshot (73)

Screenshot (68)

Screenshot (69)

Screenshot (70)

Screenshot (71)

# Risk

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "api",
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "client",
       "version": "0.1.0",
       "dependencies": {
         "@material-ui/core": "^4.12.2",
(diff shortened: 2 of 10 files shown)


sanathmahesh:
> The onboarding effect looks really nice! Just curious if there's a reason we're not implementing it to be shown automatically (by checking if a user is visiting publication page for 1st time with a flag in redux/database etc) to make the experience more 'seamless'? But I like the idea that users can go through the onboarding stuff again by clicking a button or icon. Also wondering if there will be a better way to show the '?' icon? It looks nice and neat, but just that I am afraid if our target user will notice the icon. (given that when we're doing uat the users didn't even notice the password hint icon beside the password label when registering account 😆 )

Yeah, makes sense, I will make the button a bit bigger so that it's easier for the user to see. As for making it automatically appear, that will probably be implemented in another separate PR, this PR is just adding the onboarding for now.

api/package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "api",
```
ghost: dumb question: what are the changes in package-lock.json files for?

sanathmahesh: Good catch! Not really sure how, maybe because of continuously switching between branches the package-lock files for not just api, but client and scholly service got edited and I didn't catch it in the initial commit. I fixed it now so none of those files have been changed. 👍

